PR #178, opened 2018-05-14
Issue #172: Add support for Stage and Scene as FXML file root

### LINKED ISSUE #172: SceneBuilder cannot open files with root "javafx.stage.Stage"
Originally reported by: **Yong Zhi Yuan (Bitbucket: [Zhiyuan-Amos](https://bitbucket.org/Zhiyuan-Amos), GitHub: [Zhiyuan-Amos](https://github.com/Zhiyuan-Amos))**

Some FXML files may have `Stage` as root objects to allow us to directly declare properties on the stage and scene in the FXML file.

However, SceneBuilder does not know how to handle root object of type `Stage`. I have attached a sample file as reference.

- Bitbucket: https://bitbucket.org/gluon-oss/scenebuilder/issue/172

pyokagan:
Fixes #172 

* Implement support for editing FXML files with non-Node root objects (e.g. Stages and Scenes) by introducing the concept of a "display node", which is the node to be displayed in the editor instead of the scene graph root. For a Scene, this would be the scene root, and for a Stage, this would be the scene root of its scene. This allows the contents of Stages and Scenes to be displayed in the editor instead of showing a "Content of this document cannot be displayed" error. This solution can be extended to support other non-Node objects (e.g. Tabs) in the future.

* Add support for Stage, Window and Scene attributes, and their accessories.

* To support stylesheets defined on a Scene, introduce the concept of "display stylesheets", which are the stylesheets to be applied in addition to the current theme's stylesheets.


pyokagan:
Anything I can do to move this PR forward?

abhinayagarwal:
Trying to fire a conversation here so others can chip in and share their thoughts to move this PR forward.

In my opinion, this is a really nice and long missing feature in Scene Builder. Adding support for Scene and Stage in FXML removes the burden from creating them separately. This PR is a good start to it and as [discussed earlier](https://bitbucket.org/gluon-oss/scenebuilder/pull-requests/62/issue-172-add-support-for-stage-and-scene/diff#comment-60383880), other features can be build on top of this PR.

jperedadnr:
I've tested successfully this PR, and I'd say it is almost ready to be merged. 

I have some comments, maybe to be addressed now or on a follow-up PR.

- `Scene` and `Stage` are added to `Miscellaneous` group, like `SubScene`. While that makes sense,  it is not a proper name for these "main" containers.
-  `Scene` and `Stage` don't have a proper icon (just a question mark)
- `Scene` and `Stage` don't show in the Wrap In menu, so you can't wrap a root container with a Scene or a Scene with a Stage. To do so you have to remove everything, and add the Scene or Stage in the first place. Also since neither of them are containers, maybe we could have a separator and then scene and stage.
- `Scene` fill property, combined with a transparent root, works when running the app but Scene Builder doesn't style the scene with it, neither the Preview menu.

jperedadnr:
With the latest changes it looks really good now. 

I'm not sure if you have already finished the changes for this PR. Just in case, the only thing I see is related to the preview: If you have a transparent container with a colored scene (that works now in the workspace), the preview window doesn't show it.

pyokagan:
@jperedadnr Thanks for the reviews so far.

> Scene and Stage are added to Miscellaneous group, like SubScene. While that makes sense, it is not a proper name for these "main" containers.

Do you mean we should create a new category for `Scene` and `Stage`? I'm not sure if it would be worth it to create a new category for just two items, especially since it would take up more vertical space in the Library pane. (Also, I'm not quite sure what to call this new category).

> Scene and Stage don't have a proper icon (just a question mark)

Done, by modifying existing icons. Hopefully they don't look too bad.

> Scene and Stage don't show in the Wrap In menu, so you can't wrap a root container with a Scene or a Scene with a Stage. To do so you have to remove everything, and add the Scene or Stage in the first place. Also since neither of them are containers, maybe we could have a separator and then scene and stage.

Yeah, I agree that that is a pretty crucial feature, so I've implemented it.

> Scene fill property, combined with a transparent root, works when running the app but Scene Builder doesn't style the scene with it, neither the Preview menu.

Alright, I've implemented it in the editor. The preview window is another bag of worms so I've decided to do it in another PR (along with making the preview window support the Window's minWidth, minHeight etc.)

pyokagan:
To recap, the last two remaining unimplemented features are:

* Making the preview window respect properties set on the Stage/Scene. (Currently ignored)

* Modify -> Scene Size when a Stage is the FXML root. (Currently disabled when Stage is the FXML root)

Can this PR be merged first? I'd like to see some form of support for Stage/Scenes as FXML file root in the next version of scene builder. The remaining features are a bit more involved and I'll implement them in later PRs.

PR #188, opened 2018-06-13
Add travis and appveyor ci configuration

abhinayagarwal:
Adds CI pipeline configuration to test PR and release bundles to GitHub when a new tag is pushed.


.ci/build.bat
```diff
@@ -0,0 +1,12 @@
+echo "Executing gradle"
+
+call gradlew shadowJar -PVERSION=%APPVEYOR_REPO_TAG_NAME%
```
tiainen: Do we want to explicitly run the `check` task as well here as is done in build.sh?

PR #191, opened 2018-06-21
Use native2ascii to fix garbled Japanese text

ktgw0316:
In Scene Builder Windows Installer 64-bit which is available from https://gluonhq.com/products/scene-builder/#download , all of the menu text are garbled and unreadable on Japanese locale. 'Executable Jar' seems to be mostly OK, but some of the text are still garbled.

I compared dist.jar from the Windows version and scenebuilder-8.5.0-all.jar from the Executable Jar, and found that two com/oracle/javafx/scenebuilder/app/i18n/SceneBuilderApp_ja.properties are different, so I think that the dist.jar contains wrong properties file.

I also found that the rest of the text are become readable if I replaced SceneBuilderKit_ja.properties with encoded with ```native2ascii```.


ktgw0316:
@tiainen Great! So I'll close this PR. I hope the installer for Windows will be fixed as soon as possible.

PR #196, opened 2018-09-17
fix #195 no longer use unicode escape, and specify encoding explicitl…

### LINKED ISSUE #195: Japanese characters in the main menu are garbled in executable jar built on Windows OS Japanese editin.
Regarding #194. To approach this issue,  I built from source code on Windows PC in Japanese environment.
My built executable jar shows the garbled menu text, it is different from #194 garble.

The build is the following steps:

1. git clone from scene builder repository, use branch "integration-9".
2. build with gradle (gradlew clean build)
3. run executable jar

The result shows garbled menu texts.

image

I look up the properties file for japanese in source code repository, in official released 10.0.0 executable jar(not garbled), and in my build(garbled).
To comparison easely, only look up one key and value pare of 'menu.title.file' which is shown in menu bar [File(F)] 

- In source code repository, the line is as follows. (written in UTF-8 text)

image

- In official released executable jar, the line is as follows. (written in escaped unicode text)

> menu.title.file = \u30d5\u30a1\u30a4\u30eb(F)

- In my build executable jar, the line is as follows. (written in escaped unicode text)

> menu.title.file = \u7e5d\u8f14\u3043\u7e67\uff64\u7e5d\uff6b(F)

This garble is typically occuured in Java 

by reading UTF-8 text file using java.io.FileReader API with default encoding in Windows Japanese environment.
Windows Japanese environment uses Not UTF-8 but Windows_31J encoding.

I guess this miss encoding is caused by running gradle on not UTF-8 encoding environment.

torutk:
At first, in Java SE 9, resource bundle property files are loaded in UTF-8 encoding, no longer needed unicode escaped. But there is an unicode escape process in build.gradle, so remove the line.

Next, gradle's Java plugin introduces processResources task.
This task uses gradle's Copy task internally, and the Copy task is seemed to miss encoding 
when the system encoding is different from the file copied.
So, specify encoding explicitly as UTF-8 in gradlew.bat which is used on Windows OS.

### DIFF
--- a/gradlew.bat
+++ b/gradlew.bat
@@ -9,7 +9,7 @@
 if "%OS%"=="Windows_NT" setlocal
 
 @rem Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-set DEFAULT_JVM_OPTS=
+set DEFAULT_JVM_OPTS=-Dfile.encoding=UTF-8
 
 set DIRNAME=%~dp0
 if "%DIRNAME%" == "" set DIRNAME=.
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -41,7 +41,6 @@ processResources {
             javaVersion: System.getProperty('java.runtime.version') + ', ' + System.getProperty('java.vendor'),
             buildDate: buildDate
         ])
-        filter(EscapeUnicode)
     }
     into buildDir
 }


gradlew.bat
```diff
@@ -9,7 +9,7 @@
 if "%OS%"=="Windows_NT" setlocal
 
 @rem Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-set DEFAULT_JVM_OPTS=
+set DEFAULT_JVM_OPTS=-Dfile.encoding=UTF-8
```
abhinayagarwal: It is better to add a `gradle-properties` file and pass the JVM arg in it:

```
org.gradle.jvmargs='-Dfile.encoding=UTF-8'
```

torutk: Thank you  for your good advice.
I check the gradle properties and encoding more, and will modify (or rewrite) PR.

torutk:
This pull request is for fix #195.
#195 is closed(fixed) without modification, so close this.

PR #199, opened 2018-10-25
Integration with Java 11

tiainen:
This PR allows the project to be built and run with Java 11.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,35 +1,50 @@
 import org.apache.tools.ant.filters.EscapeUnicode
 
-apply plugin: 'application'
-apply plugin: 'com.github.johnrengelman.shadow'
-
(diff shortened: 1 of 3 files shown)


jperedadnr:
It looks good, I can run it with `./gradlew run`, and also do a "light" shadow jar with `./gradlew shadowJar`, and then run with `java --module-path=/path-to/javafx-sdk-11/lib --add-modules=javafx.fxml,javafx.swing,javafx.web --add-opens=javafx.fxml/javafx.fxml=ALL-UNNAMED -jar scenebuilder-11.0.0-SNAPSHOT-all.jar`
I guess with the installer there is no need for a "fat" shadow jar that includes the javafx classes and native libraries. Anyway we might want to provide it?

PR #209, opened 2018-12-10
Store Color as an object instead of its web string representation

abhinayagarwal:
Scene Builder only supports Color as its web string representation. This largely limits what can be done with controls which have `Color` property like `ColorPicker`.

This PR enables SB to represent Color as a complex object comprising of RGBA.

With this PR, SB can represent the `ObjectProperty<Color> value` of the ColorPicker as a `PaintPicker` and can then store the FXML as:

```
<ColorPicker>
   <value>
      <Color red="0.5099999904632568" green="0.3949007987976074" blue="0.22439999878406525" opacity="0.6100000143051147" />
   </value>
</ColorPicker>
```


jperedadnr:
You didn't push the fix to the rounding issue in `com.oracle.javafx.scenebuilder.kit.util.control.paintpicker.colorpicker.ColorPicker?

PR #210, opened 2018-12-12
Wildcard import enable/disable option in preferences

### LINKED ISSUE #124: Make FXML import optimisation optional / allow or keep wildcard imports
Originally reported by: **A. Walter (Bitbucket: [awalterwork](https://bitbucket.org/awalterwork), GitHub: Unknown)**

In the resolved feature request [issue 18](https://bitbucket.org/gluon-oss/scenebuilder/issues/18/optimize-imports-in-fxml)  from 2015, it was requested that SB optimizes the FXML imports by only importing the actually used classes instead of using wildcard imports. This is now the case in **SceneBuilder 8.3.0**.

In practice, this behaviour has turned out to be very impractical for us: We want to use wildcard imports for the things we might realistically use (java.lang.\* , java.util.\* , javafx.scene.control.\* , ...) in the FXMLs so we can easily add new nodes or define objects in FXML with a text editor, but we also want to use SceneBuilder for more complex layout editing. 

We have never experienced long loading times for FXMLs that use wildcard imports, so the loading process may just have been improved since 2015. Therefore, we would like to be able to decide ourselves if we want wildcard imports or other "unnecessary" imports, instead of SceneBuilder deciding for us.

Because - every time we use SceneBuilder, all those wildcard imports are "optimized". 

This is wildly impractical if we, for example, want to temporarily comment out nodes to try layout variants (or if we are planning to add certain nodes soon after making some changes in SceneBuilder, and we have already prepared the import statements). For example:

# Steps to reproduce
* Create an FXML file with a SplitMenuButton and a Button (see attached example file) and various imports.
* Add comments around the SplitMenuButton.
* Open the file with SceneBuilder, change something (for example, the text of the remaining Button) and save the file.
* Look at the FXML file source: All imports that were not actually used (SplitMenuButton, MenuItem) are gone.
* Remove the comments around the SplitMenuButton.
* Try to open the FXML in a JavaFx application (see attached example file): It cannot be opened anymore because of the missing imports ("javafx.fxml.LoadException: SplitMenuButton is not a valid type.").
* Open the FXML file with SceneBuilder again: It will open the file, but the SplitMenuButton is not shown anywhere.

# Expected Behaviour
There should be an option to turn off FXML import optimisation so that the existing import statements are kept as-is.

Ideally, all imports already there when the file was loaded would then remain untouched by SceneBuilder, and SceneBuilder would only add (and remove, if feasible) new import statements which are not already covered by existing (wildcard) imports.

- Bitbucket: https://bitbucket.org/gluon-oss/scenebuilder/issue/124

abhinayagarwal:
Introduces a checkbox using which a user can enable whether the exported FXML when "Save/Save As" is clicked should contain wildcard imports or not.

This checkbox is unchecked by default.

Wildcards from imports was removed as a part of https://github.com/gluonhq/scenebuilder/issues/18. Since wildcard imports have a performance impact on the JavaFX application, SB will not enable it by default. Users who understand the impact of this will now have the flexibility to use it.

Fixes #124


app/src/main/java/com/oracle/javafx/scenebuilder/app/preferences/PreferencesWindowController.java
```diff
@@ -52,6 +52,7 @@
 import static com.oracle.javafx.scenebuilder.app.preferences.PreferencesController.RECENT_ITEMS;
 import static com.oracle.javafx.scenebuilder.app.preferences.PreferencesController.RECENT_ITEMS_SIZE;
 import static com.oracle.javafx.scenebuilder.app.preferences.PreferencesController.TOOL_THEME;
+import static com.oracle.javafx.scenebuilder.app.preferences.PreferencesController.WILDCARD_IMPORT;
```
jperedadnr: Where is this defined?? I don't see `WILDCARD_IMPORT` added to `PreferencesController`.

abhinayagarwal: It is defined in the `PreferencesControllerBase` class ;)

app/src/main/java/com/oracle/javafx/scenebuilder/app/preferences/PreferencesRecordGlobal.java
```diff
@@ -456,6 +466,9 @@ public void readFromJavaPreferences() {
         // Accordion animation
         setAccordionAnimation(applicationRootPreferences.getBoolean(ACCORDION_ANIMATION, true));
 
+        // Wildcard imports
+        setWildcardImports(applicationRootPreferences.getBoolean(WILDCARD_IMPORT, false));
```
tiainen: Use the field `DEFAULT_WILDCARD_IMPORTS` instead of hardcoding to `false`? The same can be done for `ACCORDION_ANIMATION` above.

kit/src/main/java/com/oracle/javafx/scenebuilder/kit/fxom/FXOMSaver.java
```diff
@@ -52,14 +51,14 @@
 class FXOMSaver {
     
     
-    public String save(FXOMDocument fxomDocument) {
+    public String save(FXOMDocument fxomDocument, boolean wildcardImports) {
```
tiainen: Can we extract `wildcardImports` to a class field and pass it to the constructor? That way, you don't need to pass the boolean to each method.

PR #217, opened 2019-01-06
Replace wildcard imports with individual class imports

abhinayagarwal:
This PR essentially cleans the code base and bring a slight performance improvement in SB by replacing wildcard imports with individual imports in all FXMLs.


abhinayagarwal:
Hi @tiainen ,

Good catch. I will give all the files another pass. 

I was also wondering if its a good idea to remove the redundant attributes from the various fxml files.

For example, `com/oracle/javafx/scenebuilder/kit/editor/panel/inspector/editors/AnchorPaneConstraintsEditor.fxml` has several instances of `prefWidth="-1.0"`.

IMO, removing these redundant attributes will help keep the fxml files a little cleaner. Thoughts?

Issue #222, opened 2019-02-02
Update build with javafx plugin


jperedadnr:
Update the build files with the javafx plugin

PR #223, opened 2019-02-02
#222 Use JavaFX gradle plugin

jperedadnr:


build.gradle
```diff
@@ -1,14 +1,8 @@
-buildscript {
-    repositories {
-        mavenCentral()
-    }
-    dependencies {
-        classpath 'com.google.gradle:osdetector-gradle-plugin:1.6.0'
-    }
+plugins {
+    id 'org.openjfx.javafxplugin' version '0.0.7'
```
tiainen: The root project probably doesn't require the plugin? We might want to add `apply false` in that case.

kit/build.gradle
```diff
@@ -1,5 +1,8 @@
 jar {
     baseName = 'scenebuilder-kit'
+    manifest {
+        attributes('Automatic-Module-Name': 'scenebuilder.kit')
```
tiainen: Can we give the module a proper name, like `com.gluonhq.scenebuilder.kit`?

kit/build.gradle
```diff
@@ -25,3 +28,10 @@ uploadArchives {
         }
     }
 }
+
+def installer = install.repositories.mavenInstaller
+def deployer = uploadArchives.repositories.mavenDeployer
+
+[installer, deployer]*.pom*.whenConfigured { pom ->
+    pom.dependencies.findAll{ it.groupId == 'org.openjfx' }.collect { it.classifier = null }
+}
```
abhinayagarwal: What are these last few lines meant to do?

tiainen: It removes the classifier from the javafx dependencies when the maven pom file for scene builder is generated. The classifier defines the platform specific part, so removing it will cause the pom to only contain platform independent javafx dependencies.

abhinayagarwal: That makes sense. Thanks for the explanation :)

app/build.gradle
```diff
@@ -1,30 +1,20 @@
...
 run {
     doFirst {
-        jvmArgs = [
-                '--module-path', classpath.asPath,
-                '--add-modules', 'javafx.fxml,javafx.swing,javafx.web,java.logging,java.scripting',
-                '--add-opens=javafx.fxml/javafx.fxml=ALL-UNNAMED'
-        ]
-        if (platform == 'linux') jvmArgs += '-Djdk.gtk.version=2'
+        jvmArgs += ['--add-opens=javafx.fxml/javafx.fxml=ALL-UNNAMED' ]
+        if (javafx.platform == 'linux') jvmArgs += '-Djdk.gtk.version=2'
```
tiainen: I missed this before, but you shoulc check the value of either `javafx.platform.classifier` or `osdetector.os`.

app/build.gradle
```diff
@@ -36,13 +26,13 @@ shadowJar {
...
 
     dependencies {
-        exclude(dependency("org.openjfx:javafx-base:11:$platform"))
-        exclude(dependency("org.openjfx:javafx-controls:11:$platform"))
-        exclude(dependency("org.openjfx:javafx-fxml:11:$platform"))
-        exclude(dependency("org.openjfx:javafx-graphics:11:$platform"))
-        exclude(dependency("org.openjfx:javafx-media:11:$platform"))
-        exclude(dependency("org.openjfx:javafx-swing:11:$platform"))
-        exclude(dependency("org.openjfx:javafx-web:11:$platform"))
+        exclude(dependency("org.openjfx:javafx-base:$javafx.version:$javafx.platform"))
```
tiainen: `$javafx.platform` should be changed to `$javafx.platform.classifier` for all the exclude directives.

PR #226, opened 2019-03-04
Update all http://nexus.gluonhq.com occurences to HTTPS

abhinayagarwal:
Updates all http nexus links with https


kit/src/main/java/com/oracle/javafx/scenebuilder/kit/editor/panel/library/maven/search/SearchService.java
```diff
@@ -112,7 +112,7 @@ protected Void call() throws Exception {
                     createSearchTask(new MavenSearch()),
                     createSearchTask(new NexusSearch(MavenPresets.SONATYPE, "http://oss.sonatype.org", "", "")),
```
JLLeitschuh: @tiainen You missed one here.

PR #228, opened 2019-04-04
UserLibrary: feature for adding folders to library

mimmoz81:
This PR adds a long waited (by me) feature, that enables the user to add a folder (in particular, root folders with classes) to the custom components library.

immagine

Users can add as many folders as they want.
After adding one folder to the library, that one is used as the root for loading classes in all its subfolders, and components that were able to be loaded are added to the "Custom" library panel, as if they loaded from a jar.

Bonus: the library label updates its text during scanning, so it is evident to users that a maybe long process is running..
immagine

The main advatange of this PR, is that a developer can add a "bin" folder from their running Eclipse/IDEA to the library manager and have the custom components loaded into SB without the need to create and publish a jar to the library directory everytime.

Feel free to ask for clarifications...

### DIFF
--- a/kit/src/main/java/com/oracle/javafx/scenebuilder/kit/editor/panel/library/ImportWindowController.java
+++ b/kit/src/main/java/com/oracle/javafx/scenebuilder/kit/editor/panel/library/ImportWindowController.java
@@ -595,7 +595,13 @@ public class ImportWindowController extends AbstractModalDialog {
         try {
             int index = 0;
             for (File file : files) {
-                result[index] = new URL("jar","",file.toURI().toURL()+"!/");
+            	URL url = file.toURI().toURL();
+                if (url.toString().endsWith(".jar")) {
+                    result[index] = new URL("jar", "", url + "!/"); // <-- jar:file/path/to/jar!/
+                } else {
+                    result[index] = url; // <-- file:/path/to/folder/
+                }
+                
                 index++;
             }
         } catch (MalformedURLException x) {
--- a/kit/src/main/java/com/oracle/javafx/scenebuilder/kit/editor/panel/library/ImportWindowController.java
+++ b/kit/src/main/java/com/oracle/javafx/scenebuilder/kit/editor/panel/library/ImportWindowController.java
@@ -595,13 +595,13 @@ public class ImportWindowController extends AbstractModalDialog {
         try {
             int index = 0;
             for (File file : files) {
-            	URL url = file.toURI().toURL();
+                URL url = file.toURI().toURL();
                 if (url.toString().endsWith(".jar")) {
                     result[index] = new URL("jar", "", url + "!/"); // <-- jar:file/path/to/jar!/
                 } else {
                     result[index] = url; // <-- file:/path/to/folder/
                 }
-                
+
                 index++;
             }
         } catch (MalformedURLException x) {
--- a/kit/src/main/java/com/oracle/javafx/scenebuilder/kit/editor/panel/library/ImportWindowController.java
+++ b/kit/src/main/java/com/oracle/javafx/scenebuilder/kit/editor/panel/library/ImportWindowController.java
@@ -32,16 +32,6 @@
  */
 package com.oracle.javafx.scenebuilder.kit.editor.panel.library;
 
-import com.oracle.javafx.scenebuilder.kit.alert.ImportingGluonControlsAlert;
-import com.oracle.javafx.scenebuilder.kit.i18n.I18N;
-import com.oracle.javafx.scenebuilder.kit.editor.panel.util.dialog.AbstractModalDialog;
-import com.oracle.javafx.scenebuilder.kit.editor.panel.util.dialog.ErrorDialog;
-import com.oracle.javafx.scenebuilder.kit.fxom.FXOMDocument;
-import com.oracle.javafx.scenebuilder.kit.library.user.UserLibrary;
-import com.oracle.javafx.scenebuilder.kit.library.util.JarExplorer;
-import com.oracle.javafx.scenebuilder.kit.library.util.JarReport;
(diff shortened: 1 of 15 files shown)


mimmoz81:
Just a couple of implementation notes:

1. class FolderExplorer was nearly cloned from JarExplorer, reusing JarReport and JarReportEntry: that was done in order to minizime the changes, but it would be better to create 2 base classes LibraryReport and LibraryReportEntry and have JarReport/Entry and FolderReport/Entry extends those bases classes

2. Because of 1. i create LibraryUtil class, just to have some common methods inside it.. having the base classes as of 1., LibraryUtil is maybe useless.

Tell me if you want me to make some change. (anyway, I noticed that LibraryUtil is missing the license header...)

jperedadnr:
This looks like a nice feature to be added to Scene Builder. Thanks for your PR!

I've tried it but I had some issues to make it work.

- If you provide a folder that has a jar in it, it won't work, as it will explore only `.class` files, but won't explore the jar itself. Is this correct? Maybe this can be fixed.

- If you provide a folder (bin, out, build...) that contain `.class` files, it doesn't work for me either. This was unexpected.

If you check this method: `LibraryFolderWatcher::makeURLArrayFromPaths`, the class path is configured only with jars:

```
result[i++] = new URL("jar", "", p.toUri().toURL() + "!/");
``
so when you add your folder (bin, out, build, ...) to the class path, it will include a jar that doesn't exist.
So I had to change it to make it work to:

```
try {
                URL url = p.toUri().toURL();
                if (url.toString().endsWith(".jar")) {
                    result[i++] = new URL("jar", "", url + "!/"); // <-- jar:file/path/to/jar!/
                } else {
                    result[i++] = url; // <-- file:/path/to/folder/
                }
            } ... 
```
That loads any jar and also any folder with the usual file protocol.

The same applies to `ImportWindowController::makeURLArrayFromFiles`. Otherwise, the default Import Dialog won't find the control, and will show "Import Jar (no UI components)" caption.  

Next issue: You need to provide the exact folder that has the classes based on their package name.

For instance, for a control under `/path/to/out/production/com/mycompany/control.class`

- this doesn't work: `/path/to/out/`
- but this works: `/path/to/out/production`.

While this is okay, somehow it should be advised. Or ideally, some kind of recursion algorithm could be used.

Another possible issue: what happens with dependencies? When we import a control from a jar, all traverse dependencies are resolved, but how does it apply in this case?

**Comment by mimmoz81** [comment]
Hello Josè, thanks for the reply.

Just a quick reply (I'm in a hurry within a meeting..):

Exploring jars inside the specified folder wasn't supposed to work: just .classes. Probably you are right it could be interesting, but I would add this functionality with anoter PR; My aim is just to add capability to work with the binaries compiled by IDEs.

Let me check in more detail the dependencies issue. I will reply ASAP.

bye

**Comment by mimmoz81** [comment]
Well, I've tested current dependency discovery feature for directory-based library and jar-based, and have found that it almost works as expected. I have a test case that fails to load a custom component (using SB 8.5.0), but I'll post an issue for that (this affects both jar-based library components both directory-based).

Can you elaborated better, please?

In the meanwhile, i'm going to push the fixes you suggested (and thanks for that).

**Comment by mimmoz81** [comment]
Do you think this text should be better?
immagine

**Comment by mimmoz81** [comment]
@eduarddrenth maybe you could be interested in this?

**Comment by FilloFolla** [comment]
Great @mimmoz81. It's a good and useful work!

**Comment by mimmoz81** [comment]
@jperedadnr any news?
do you want me to do something else about this?

**Comment by jperedadnr** [comment]
Maybe we should try to shorten the text, something like:
> Add root folder with *.class files 
And maybe you can add a tooltip for a more verbose explanation?

**Comment by mimmoz81** [comment]
Here you are
immagine

File: kit/src/main/java/com/oracle/javafx/scenebuilder/kit/editor/panel/library/ImportWindowController.java [hunk 1]
```diff
@@ -228,10 +231,36 @@ protected void okButtonPressed(ActionEvent e) {
...
+            
             if (copyFilesToUserLibraryDir) {
-                libPanelController.copyFilesToUserLibraryDir(importFiles);
+            	// collect directories from importFiles and add to library.folders file
+            	// for other filex (jar, fxml) copy them directly
+            	List<File> folders = new ArrayList<>(importFiles.size());
+            	List<File> files = new ArrayList<>(importFiles.size());
+            	
+            	for (File file : importFiles) {
+					if (file.isDirectory())
```
jperedadnr: Can you fix these indenting issues?

mimmoz81: done

jperedadnr:
Thanks, before we can merge this PR, would you mind signing this [CLA](https://docs.google.com/forms/d/16aoFTmzs8lZTfiyrEm8YgMqMYaGQl0J8wA0VJE2LCCY)?

mimmoz81:
Signed

jperedadnr:
@mimmoz81 Hey, as we have already merged this PR, I've started with a new one, that will provide logging to report the result of importing custom controls from jar or folders: #234 

I've started also some refactoring, so feel free to review and test.

mimmoz81:
Hello, thanks for merging this!

I'm preparing other useful stuffs about custom controls: I'll do another PR ASAP; if you are planning a release soon, please tell me when so that this new PR will get merged too.

In the meanwhile, I'll check your modifications.

UltraBurstXD:
Why is this PR not available for Linux Scene Builder too?

mimmoz81:
Do you mean it does not work on Linux machine or it is simply missing from binary release?

UltraBurstXD:
It's missing.

PR #229, opened 2019-04-12
MigPane integration & support improved for custom components.

dokshinit:
Integration for MigPane layout is implemented:
* MigPane component is added to the list of built-in containers.
* Properties for layout and child constraints are added to the inspector panel.

Support for custom components is improved:
* Correct interaction is provided for custom components that are context-oriented and not inherited from Control.
* The propagation of the static property relevance is provided to the descendants of the target container class.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,6 +33,10 @@ subprojects {
     }
 
     dependencies {
+        // MigPane integration.
+        compile 'com.miglayout:miglayout-core:5.2'
+        compile 'com.miglayout:miglayout-javafx:5.2'
+
         // maven dependencies
         compile 'org.eclipse.aether:aether-api:1.1.0'
         compile 'org.eclipse.aether:aether-impl:1.1.0'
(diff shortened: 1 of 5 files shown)


Oliver-Loeffler:
@dokshinit would you retrofit this PR to work with SB16? Its quite a while pending and, I think it could be a good thing.

@abhinayagarwal 
What do you think?

abhinayagarwal:
MigPane is not a core control of JavaFX which make me hesitant to bake it into Scene Builder.

Oliver-Loeffler:
Okay, this is an important point to be mentioned in CONTRIBUTING.md.

I understand this, as when issues with such components occur, the place to start an issue is most likely not Scene Builder at Github. There are a few options I see.
* not use external controls at all - with having a statement on that in contribution rules
* provide external components under some restrictions and rules, such the handling of issues is clarified etc.
* update MigPane so that it integrates well as an external control in scene builder, so that it can be use via existing SceneBuilder repositories - this would have the advantage, that also newer or older versions can be used.

Oliver-Loeffler:
@AlmasB If the guideline is to not include external libraries in Scene Builder as of now, we should review this request to check if certain parts are beneficial for Scene Builder and close this. Nevertheless, this is a candidate for a potential plugins project.

AlmasB:
As this is external to JavaFX core, it shouldn't go into `app` or `kit`. This is indeed a good candidate for the later `plugins` module, which is why this is left as `DRAFT` until then.

PR #241, opened 2019-05-13
DRAFT: Remove Charm dependencies from kit

TobiasAsk:
It would be nice to make the _kit_ project independent of the Charm Glisten dependency. Even though it is available at the Nexus repository, it is not open source as far as I can tell, and thus also not redistributable. This PR is a first step towards extracting the dependency out of the _kit_ project, and into a new subproject, on which the _app_ project depends. It is not a complete solution yet, as I am interested in hearing your feedback before going further with it.

The BuiltinLibrary's dependency has been removed here, in the following way: The part of BuiltinLibrary's constructor that adds the custom components is changed to acquire them from a component provider instead. The discovery of providers is extracted to a protected method to make the library extendable. In particluar, this enables the Eclipse integration to acquire providers through its extension point mechanism. In BuiltinLibrary, providers are discovered using Java's builtin SPI mechanism.


AlmasB:
Thanks for this. I agree with non-core JavaFX dependencies not being inside `kit`. There is a future plan for `plugins`, (seems to be called components in this PR, but I think the concept is the same), so until then this is parked as `DRAFT`.

PR #261, opened 2019-12-06
Gradle Java 14 compatibility 

garzy:
gradlew build instructions doesn't work with java 13 (says incompatible version 57). Also, shadow plugin version is very old and also says that it's incompatible with java13.

I've upgrade gradlew, shadow plugin and javafx plugin to latest versions and it's works


abhinayagarwal:
Can you update Gradle version to 6.3 which is compatible with Java 14?

PR #271, opened 2020-03-20
Fixed Japanese garbled characters 

logue:
In addition, the translation and some notations of the untranslated parts have been corrected.
image

### DIFF
--- a/app/src/main/resources/com/oracle/javafx/scenebuilder/app/i18n/SceneBuilderApp_ja.properties
+++ b/app/src/main/resources/com/oracle/javafx/scenebuilder/app/i18n/SceneBuilderApp_ja.properties
@@ -1,4 +1,4 @@
-# Copyright (c) 2016, 2018, Gluon and/or its affiliates.
+# Copyright (c) 2016, 2019, Gluon and/or its affiliates.
 # Copyright (c) 2012, 2014, Oracle and/or its affiliates.
 # All rights reserved. Use is subject to license terms.
 #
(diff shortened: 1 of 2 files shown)


abhinayagarwal:
Is there an issue logged for this PR? This used to work in the past.

logue:
The distributed binaries were garbled. Also, when I built from the latest version of the source on Github, it was still garbled.

Probably due to not running ascii2native.

The property file recognizes only ISO-8859-1. For this reason, UTF-8 encoded text cannot be read correctly, and I think it is garbled.

In addition, it does not seem to occur when building with Java 9 or later.

> This used to work in the past.

Perhaps it wasn't garbled because the properties file was Shift_JIS?
With this method, the source code may be garbled, and problems may occur when creating multilingual applications.

abhinayagarwal:
@logue Thanks for the clarification.

Default file encoding for ResourceBundle class to load properties files was changed from ISO-8859-1 to UTF-8 under [JEP-226](https://openjdk.java.net/jeps/226) and is available from JDK 9 onward.

All future releases for Scene Builder will use JDK 11 or later, which makes this PR redundant. Thoughts?

logue:
Yes.
I think that there is no problem if you build with Java 9 or later (actually Java 11 LTS). 

Maybe add the following line to settings.gradle:
```
sourceCompatibility = 11
targetCompatibility = 11
```

This pull request also includes corrections for untranslated or mistranslated parts.
In this case, please reflect only the 88d3b7d8f0e5db6727e8250524d71137edd5ac41 commit.

abhinayagarwal:
Could you update this PR or create a new PR for corrections for untranslated or mistranslated parts?

I will have an internal discussion and take care of:

```
sourceCompatibility = 11
targetCompatibility = 11
```

abhinayagarwal:
Closing as inactive

PR #277, opened 2020-06-03
section name for imported libraries

safirex:
A minor update changing the section names of imported librairies for more visibility

from  
image
to  
image

### DIFF
--- a/kit/src/main/java/com/oracle/javafx/scenebuilder/kit/library/user/LibraryFolderWatcher.java
+++ b/kit/src/main/java/com/oracle/javafx/scenebuilder/kit/library/user/LibraryFolderWatcher.java
@@ -411,9 +411,19 @@ class LibraryFolderWatcher implements Runnable {
                 final String canonicalName = e.getKlass().getCanonicalName();
                 if (!excludedItems.contains(canonicalName) && 
                     !artifactsFilter.contains(canonicalName)) {
-                    final String name = e.getKlass().getSimpleName();
-                    final String fxmlText = JarExplorer.makeFxmlText(e.getKlass());
-                    result.add(new LibraryItem(name, UserLibrary.TAG_USER_DEFINED, fxmlText, iconURL, library));
+                	
+                	final String name = e.getKlass().getSimpleName();
+                	String sectionName =jarOrFolderReport.getJar().toString();
+                	
+                	// if some os don't use '/' for folder path (don't know if there is)
+                	if (sectionName.lastIndexOf("\\")==-1)
+                		sectionName=UserLibrary.TAG_USER_DEFINED;
+                	else
+                		sectionName=sectionName.substring(sectionName.lastIndexOf("\\")+1, 
+                					sectionName.lastIndexOf("."));
+                	
+                	final String fxmlText = JarExplorer.makeFxmlText(e.getKlass());
+                    result.add(new LibraryItem(name, sectionName, fxmlText, iconURL, library));
                 }
             }
         }


kit/src/main/java/com/oracle/javafx/scenebuilder/kit/library/user/LibraryFolderWatcher.java
```diff
@@ -411,9 +411,19 @@ else if (Files.isDirectory(currentJarOrFolder)) {
...
-                    final String fxmlText = JarExplorer.makeFxmlText(e.getKlass());
-                    result.add(new LibraryItem(name, UserLibrary.TAG_USER_DEFINED, fxmlText, iconURL, library));
+                	
+                	final String name = e.getKlass().getSimpleName();
+                	String sectionName =jarOrFolderReport.getJar().toString();
+                	
+                	// if some os don't use '/' for folder path (don't know if there is)
+                	if (sectionName.lastIndexOf("\\")==-1)
+                		sectionName=UserLibrary.TAG_USER_DEFINED;
+                	else
```
johanvos, line 421: best to use {} here in case other statements are added

kit/src/main/java/com/oracle/javafx/scenebuilder/kit/library/user/LibraryFolderWatcher.java
```diff
@@ -411,9 +411,19 @@ else if (Files.isDirectory(currentJarOrFolder)) {
                 final String canonicalName = e.getKlass().getCanonicalName();
                 if (!excludedItems.contains(canonicalName) && 
                     !artifactsFilter.contains(canonicalName)) {
-                    final String name = e.getKlass().getSimpleName();
-                    final String fxmlText = JarExplorer.makeFxmlText(e.getKlass());
-                    result.add(new LibraryItem(name, UserLibrary.TAG_USER_DEFINED, fxmlText, iconURL, library));
+                	
+                	final String name = e.getKlass().getSimpleName();
+                	String sectionName =jarOrFolderReport.getJar().toString();
```
johanvos, line 416: are you sure getJar() can not return null?

kit/src/main/java/com/oracle/javafx/scenebuilder/kit/library/user/LibraryFolderWatcher.java
```diff
@@ -411,9 +411,19 @@ else if (Files.isDirectory(currentJarOrFolder)) {
...
+                	
+                	// if some os don't use '/' for folder path (don't know if there is)
+                	if (sectionName.lastIndexOf("\\")==-1)
+                		sectionName=UserLibrary.TAG_USER_DEFINED;
+                	else
+                		sectionName=sectionName.substring(sectionName.lastIndexOf("\\")+1, 
+                					sectionName.lastIndexOf("."));
```
johanvos, line 423: what if there is no `.` in the name, or it occurs before the last `'` . This should not happen, but in that case, it makes sense to fallback to the default?

kit/src/main/java/com/oracle/javafx/scenebuilder/kit/library/user/LibraryFolderWatcher.java
```diff
@@ -411,9 +411,19 @@ else if (Files.isDirectory(currentJarOrFolder)) {
...
+                	
+                	final String fxmlText = JarExplorer.makeFxmlText(e.getKlass());
```
johanvos, line 425: replace tab with spaces

abhinayagarwal:
Hi @safirex ,

Are you still willing to work on this PR?

safirex:
@abhinayagarwal 
hello,
not at the moment, no.

kit/src/main/java/com/oracle/javafx/scenebuilder/kit/library/user/LibraryFolderWatcher.java
```diff
@@ -438,8 +438,15 @@ else if (Files.isDirectory(currentJarOrFolder)) {
                 if (!excludedItems.contains(canonicalName) && 
                     !artifactsFilter.contains(canonicalName)) {
                     final String name = e.getKlass().getSimpleName();
+                    String sectionName;
+                    try{
```
abhinayagarwal, line 442: Please add a space between try and braces

kit/src/main/java/com/oracle/javafx/scenebuilder/kit/library/user/LibraryFolderWatcher.java
```diff
@@ -438,8 +438,15 @@ else if (Files.isDirectory(currentJarOrFolder)) {
...
+                        sectionName=jarOrFolderReport.getJar().toFile().getName().replace(".jar", "");
+                    }catch(Exception exception){
```
abhinayagarwal, line 444: Please add a space catch and each of the braces

kit/src/main/java/com/oracle/javafx/scenebuilder/kit/library/user/LibraryFolderWatcher.java
```diff
@@ -438,8 +438,15 @@ else if (Files.isDirectory(currentJarOrFolder)) {
...
+                        //if getJar returns null or toFile not finding path
+                        sectionName=UserLibrary.TAG_USER_DEFINED;
+                    }
```
abhinayagarwal, line 447: I would move the `try..catch` block into a separate private method: 

```
String sectionName = createSectionName();
...

private String createSectionName() {
    try {
        return jarOrFolderReport.getJar().toFile().getName().replace(".jar", "");
    } catch (Exception e) {
        return UserLibrary.TAG_USER_DEFINED;
    }
}
```

abhinayagarwal:
Hi @safirex ,

I have left some comments.

I couldn't find you in our contributors list. Can you please sign the [Gluon Individual Contributor License Agreement](https://cla.gluonhq.com).

abhinayagarwal:
Looks good. One more change which I missed to point out earlier. Please update the license header on the file to current year:


Copyright (c) 2017, ~2020~2021, Gluon and/or its affiliates

abhinayagarwal (approved):

kit/src/main/java/com/oracle/javafx/scenebuilder/kit/library/user/LibraryFolderWatcher.java
```diff
@@ -438,15 +438,22 @@ else if (Files.isDirectory(currentJarOrFolder)) {
                 if (!excludedItems.contains(canonicalName) && 
                     !artifactsFilter.contains(canonicalName)) {
                     final String name = e.getKlass().getSimpleName();
+                    final String sectionName = createSectionName(jarOrFolderReport);
```
jperedadnr, line 441: This should be outside the for loop

abhinayagarwal: Good catch, José!

jperedadnr:
In general this PR looks good, providing a section name to categorise all different custom controls definitely would help.

However, if the user happens to have many imported jars, the list grows at the top level, and the number of root categories in the accordion is too big, causing the titled panes not to show any content or barely one or two items of a possibly long list. Also, this will clutter the view, having the built-in categories (like Controls) at the end and barely accessible.

To move this PR forward, we need somehow to introduce the concept of sub-category.

As the library items are currently displayed in a titled pane, using a single `ListView<LibraryListItem>`, I suggest a few options:
- Use the subcategory title as a Label node, having a ListView control for each jar. 
- Create a LibraryListItem to be used as header, for each subcategory, so a single listView will do, however this will require two different custom cells, and figuring out how to sort it (first by sub-category, then by name).
- Based on the above, the `CharmListView` control is already designed for that. The required refactoring is fairly easy to do.

If we keep the accordion as is (with each custom library as a top categories and no sub-categories), we could add a button to the right of the Search field, so the user can manage the categories, hiding the ones not frequently used.

Issue #319, opened 2021-02-01
Included fxml component can't be focused.


bdh92123:
Any fxml control included using 'FIle-Include-FXML' can't be focused.
To focus it, It should be wrapped in AnchorPane (Before include it, focus any anchor pane). 
But even so, there is a problem that the right-click context menu does not appear.

### Expected Behavior
Fxml control included can be focusd and resize, move, etc. 

### Steps to Reproduce
Include other fxml in hierarchy using 'FIle-Include-FXML'.

### Your Environment
Java
15.0.2+7, AdoptOpenJDK

Operating System
Windows 10, amd64, 10.0

abhinayagarwal:
Hi @bdh92123 ,

Can you please provide a sample FXML to reproduce the issue?

bdh92123:
Hi @abhinayagarwal .
The following shows the reproduction of the problem and after applying [PR #321](url).

Before
![bug](https://user-images.githubusercontent.com/2897088/108031061-8a5f8d80-7073-11eb-967a-7228e55937b1.gif)

After
![fix](https://user-images.githubusercontent.com/2897088/108031068-8cc1e780-7073-11eb-8b53-5cecafb3d7c2.gif)

And comp.fxml in screenshot is following:
```<?xml version="1.0" encoding="UTF-8"?>

<?import javafx.scene.control.Button?>
<?import javafx.scene.control.TextField?>
<?import javafx.scene.layout.HBox?>

<HBox maxHeight="-Infinity" maxWidth="-Infinity" minHeight="-Infinity" minWidth="-Infinity" xmlns="http://javafx.com/javafx/15.0.1" xmlns:fx="http://javafx.com/fxml/1">
   <children>
      <TextField HBox.hgrow="ALWAYS" />
      <Button maxHeight="-Infinity" maxWidth="-Infinity" minHeight="-Infinity" minWidth="-Infinity" mnemonicParsing="false" prefHeight="22.0" prefWidth="57.0" text="Button" />
   </children>
</HBox>
```

Thanks.

PR #321, opened 2021-02-04
Included fxml should can be handled as node.

bdh92123:
### Issue
https://github.com/gluonhq/scenebuilder/issues/319
Included fxml can't be focused, moved, resized.

### Fixes
- Included fxml also can be handled in scene builder. So it can also be focused, moved or resized like other node.
- Fixed a bug where the context menu did not appear on included fxml nodes.


johanvos:
Interesting.
It would be good to have a test that fails before and runs after this patch is applied, can you add some test for this?

bdh92123:
Hi @johanvos . I uploaded some test in Issue #319.
Thanks.

abhinayagarwal:
Hi @bdh92123,

Thank you for the PR. How big of an effort is it to properly fix this in the Intrinsic node?

bdh92123:
Hi @abhinayagarwal ,
Scene Builder is designed so that only nodes with bounds (FXOMInstance) can be positioned or resized. Intrinsic node is treated as a completely different class (FXIntrinsic), so if you don't use the trick I did, I think it may require quite a bit of effort.
Thanks.

Issue #354, opened 2021-04-09
SceneBuilder doesn't open or it's really slow in IntelliJ


palexdev:
### Expected Behavior
When I open SceneBuilder from any shortcut it opens almost immediately.

### Current Behavior
In IntelliJ, it requires some time to open, it's really slow.

### Steps to Reproduce

1. Right-click on a fxml file
2. Click on "Open In SceneBuilder"

### Your Environment
OS: Windows 10 20H2 x64, SceneBuilder v16.0.0, IntelliJ IDEA 2021.1

### Additional Details
I submitted an issue on JetBrain's YourTrack and they said that it is a third-party problem.
Link to the issue: [YourTrack](https://youtrack.jetbrains.com/issue/IDEA-266297)

abhinayagarwal:
Hi @palexdev !

Thank you for opening this issue.

Do you have the same experience in NB and Eclipse?

palexdev:
Just tested on a VM, both NB and Eclipse work fine. SceneBuilder opens immediately.
I was wondering if the problem could be the javafx plugin for intellij at this point

palexdev:
@abhinayagarwal is it possible to enable logging or is it already enabled?
I noticed that if I restart the process with a process manager SceneBuilder opens immediately (the files is loaded too)
Could it be some JVM exception that slows it down?

palexdev:
https://user-images.githubusercontent.com/16880178/125312560-ad495180-e334-11eb-8688-eb394a5cc99e.mp4

Video of the actual behavior

palexdev:
Closing as fixed by #358
But at the time of writing this, the latest SceneBuilder build available still doesn't have the fix, so you either wait for the new version or compile it by yourself (good luck with that)

PR #371, opened 2021-07-09
Load logging configuration in Application class

### LINKED ISSUE #370: Log files are not created
The About dialog mentions that the log file must be created in tmp directory. However, no log files are created.

This is because no log file config is loaded by Scene Builder.

### Expected Behavior

Log file should be generated when Scene Builder runs.

### Current Behavior

No log files are generated

### Steps to Reproduce

Run Scene Builder on any platform and no log files are created.

abhinayagarwal:
### Issue

Fixes #370 

### Progress

### DIFF
--- a/app/src/main/java/com/oracle/javafx/scenebuilder/app/AppPlatform.java
+++ b/app/src/main/java/com/oracle/javafx/scenebuilder/app/AppPlatform.java
@@ -54,6 +54,7 @@ public class AppPlatform {
     private static String applicationDataFolder;
     private static String userLibraryFolder;
     private static String messageBoxFolder;
+    private static String logsFolder;
     private static MessageBox<MessageBoxMessage> messageBox;
     
     public static synchronized String getApplicationDataFolder() {
@@ -90,6 +91,17 @@ public class AppPlatform {
         return userLibraryFolder;
     }
 
+    /**
+     * Returns the directory path for logs. Default path is "${user.home}/.scenebuilder/logs/".
+     * @return Directory path for Scene Builder logs
+     */
+    public static synchronized String getLogFolder() {
+        if (logsFolder == null) {
+            logsFolder = getApplicationDataFolder() + "/logs"; //NOI18N
+        }
+        return logsFolder;
+    }
+
     public static boolean requestStart(
             AppNotificationHandler notificationHandler, Application.Parameters parameters)  
     throws IOException {
@@ -123,6 +135,7 @@ public class AppPlatform {
         
         try {
             Files.createDirectories(Paths.get(getMessageBoxFolder()));
+            Files.createDirectories(Paths.get(getLogFolder()));
         } catch(FileAlreadyExistsException x) {
             // Fine
         }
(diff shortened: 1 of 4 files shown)


app/src/main/resources/logging.properties
```diff
@@ -0,0 +1,7 @@
+handlers= java.util.logging.FileHandler
+
+# File Logging
+java.util.logging.FileHandler.pattern = ${java.io.tmpdir}/scenebuilder-${version}.log
```
jperedadnr, line 4: Probably better to log to `~/.scenebuilder` (and change the message on the dialog)? It would be more permanent.

palexdev:
Yay finally, with this we can probably investigate further Issue #354 (Updated with video)

app/src/main/resources/logging.properties
```diff
@@ -0,0 +1,7 @@
+handlers= java.util.logging.FileHandler
+
+# File Logging
+java.util.logging.FileHandler.pattern = %h/.scenebuilder/logs/scenebuilder-${version}.log
```
jperedadnr, line 4: this fails on Mac:
```
Can't load log handler "java.util.logging.FileHandler"
java.nio.file.NoSuchFileException: /Users/JosePereda/.scenebuilder/logs/scenebuilder-16.0.0-SNAPSHOT.log.lck
```

If you check AppPlatform, the app folder is not `~/.scenebuilder` for all platforms.

abhinayagarwal: I now see that the [Application Data directory](https://github.com/gluonhq/scenebuilder/blob/master/app/src/main/java/com/oracle/javafx/scenebuilder/app/AppPlatform.java#L64) is different for different platforms. I wonder why was it done instead of using `${user.home}/.scenebuilder`?

jperedadnr: Can't reply to that :) 
If needed for logs only you might want to simply use `${user.home}/.scenebuilder` for all platforms?

app/src/main/java/com/oracle/javafx/scenebuilder/app/AppPlatform.java
```diff
@@ -90,6 +91,17 @@ public static synchronized String getUserLibraryFolder() {
         return userLibraryFolder;
     }
 
+    /**
+     * Returns the directory path for logs. Default path is "${user.home}/.scenebuilder/logs/".
+     * @return Directory path for Scene Builder logs
+     */
+    public static synchronized String getLogFolder() {
+        if (logsFolder == null) {
+            logsFolder = System.getProperty("user.home") + "/.scenebuilder/logs"; //NOI18N
```
jperedadnr, line 100: Does this work on Windows? `File.separator` is probably needed

app/src/main/resources/logging.properties
```diff
@@ -0,0 +1,7 @@
...
+java.util.logging.FileHandler.count = 1
```
jperedadnr, line 5: I see that the log is restarted whenever you stop and start a new instance of Scene Builder, removing the previous logs.
Should we just have a rolling system with a maximum number of files (like 10), and size per file (like 1 MB), but keep everything until the limit is reached?

jperedadnr (approved):

credmond:
Hi @abhinayagarwal, this change introduced a bug whereby no logging file ever gets created when SceneBuilder was ran from a JAR. I have fixed this here:

https://github.com/gluonhq/scenebuilder/pull/658

... given the amount if bugs in SceneBuilder, I think logging is a priority. Can we get this into the next release please? Reproducing is easy, just run SceneBuilder from a JAR or from the packaged SceneBuilder.exe.

PR #372, opened 2021-07-13
Skip update check in developer mode

### LINKED ISSUE #352: Popup with failed Update Check in Dev Mode
When starting SceneBuilder locally from master an error popup for a failed "Update Check" is shown.
It states in the text that this is probably due to "Version number format not supported. Maybe using SNAPSHOT or RC versions".
If starting SceneBuilder frequently this gets annoying.
Additionally, the main window size changes and is smaller when the popup shows up.

### Expected Behavior
No error popup is shown. The main window size is not changed.

### Current Behavior
An error popup "Update check" is shown. The main window size is reduced.

### Steps to Reproduce
1. Start SceneBuilder with `mvn javafx:run -f app`.
2. Close the window with the recent FXML files opened (title "Gluon Scene Builder").
3. Error popup should appear. Size of the main window is changed.
 
### Your Environment
SceneBuilder built from master
Linux

### Screenshots
scenebuilder_update_check

abhinayagarwal:
Fixes #352

### DIFF
--- a/app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
+++ b/app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
@@ -939,7 +939,9 @@ public class SceneBuilderApp extends Application implements AppPlatform.AppNotif
             }
             try {
                 boolean showUpdateDialog = true;
-                if (AppSettings.isCurrentVersionLowerThan(latestVersion)) {
+                if (AppSettings.getSceneBuilderVersion().contains("SNAPSHOT")) {
+                    showUpdateDialog = false;
+                } else if (AppSettings.isCurrentVersionLowerThan(latestVersion)) {
                     PreferencesController pc = PreferencesController.getSingleton();
                     PreferencesRecordGlobal recordGlobal = pc.getRecordGlobal();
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
```diff
@@ -1011,7 +1013,7 @@ private void showVersionNumberFormatError(DocumentWindowController dwc) {
         // in development so we don't localize the strings
         alert.setTitle("Error");
         alert.setHeaderText(I18N.getString("check_for_updates.alert.headertext"));
-        alert.setContentText("Version number format not supported. Maybe using SNAPSHOT or RC versions.");
+        alert.setContentText("Update check is disabled in development environment.");
```
jperedadnr, line 1016: The new snapshot check doesn't throw NFE. If it is not a snapshot, and while checking the version there is a NFE, the error message should still be `Version number format not supported`.

abhinayagarwal: The new snapshot will not throw NFE during startup. However, it will throw if a user goes selects "Check for updates.." from the Help menu.

jperedadnr (approved):

PR #387, opened 2021-08-11
Fix for issue #386

### LINKED ISSUE #386: FileSystemWatchService is never stopped and is spawned multiple times
The watchService for user library is never stopped and it is spawned twice or more.
I'm creating this issue as described in #236 
I'm going to do a PR for this shortly (I hope..)

### Expected Behavior
The watchService should be terminated gracefully

### Current Behavior
The watchService is never stopped

### Steps to Reproduce
Import a library from jar/fxml manager.
Drop that library.
Install again.
Drop again.
Install again.

image

You will have many threads named "FileSystemWatchService" doing the same job and not releasing resources.

image

mimmoz81:
This PR fixes #386.
FileSystemWatcherService is stopped on a finally block.

### DIFF
--- a/kit/src/main/java/com/oracle/javafx/scenebuilder/kit/library/user/LibraryFolderWatcher.java
+++ b/kit/src/main/java/com/oracle/javafx/scenebuilder/kit/library/user/LibraryFolderWatcher.java
@@ -161,98 +161,110 @@ class LibraryFolderWatcher implements Runnable {
     }
     
     private void runWatching() throws InterruptedException {
-        while (true) {
-            final Path folder = Paths.get(library.getPath());
+        WatchService watchService = null;
+        try {
+            while (true) {
+                final Path folder = Paths.get(library.getPath());
 
-            WatchService watchService = null;
-            while (watchService == null) {
-                try {
-                    watchService = folder.getFileSystem().newWatchService();
-                } catch(IOException x) {
-                    System.out.println("FileSystem.newWatchService() failed"); //NOI18N
-                    System.out.println("Sleeping..."); //NOI18N
-                    Thread.sleep(1000 /* ms */);
+                while (watchService == null) {
+                    try {
+                        watchService = folder.getFileSystem().newWatchService();
+                    } catch(IOException x) {
+                        System.out.println("FileSystem.newWatchService() failed"); //NOI18N
+                        System.out.println("Sleeping..."); //NOI18N
+                        Thread.sleep(1000 /* ms */);
+                    }
                 }
-            }
 
-            WatchKey watchKey = null;
-            while ((watchKey == null) || (watchKey.isValid() == false)) {
-                try {
-                    watchKey = folder.register(watchService, 
-                    StandardWatchEventKinds.ENTRY_CREATE, 
-                    StandardWatchEventKinds.ENTRY_DELETE, 
-                    StandardWatchEventKinds.ENTRY_MODIFY);
+                WatchKey watchKey = null;
+                while ((watchKey == null) || (watchKey.isValid() == false)) {
+                    try {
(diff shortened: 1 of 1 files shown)


abhinayagarwal:
This is a simple change and looks OK to me. I would also like @jperedadnr to have a look a look at it before we go ahead and merge the same.

kit/src/main/java/com/oracle/javafx/scenebuilder/kit/library/user/LibraryFolderWatcher.java
```diff
@@ -161,98 +161,110 @@ private void runDiscovery() throws InterruptedException {
...
+                        } while (wk.reset());
+                    } catch(IOException x) {
+                        Thread.sleep(1000 /* ms */);
+                    }
                 }
-            }
 
+            }
+        }
+        finally {
```
jperedadnr, line 259: Do you think it would make sense to move the `InterruptException` thrown from the method signature to a `catch` in the new  `try {} finally {}` ?
I don't think there is any difference, but it might help clarifying how the `while (true) {}` is terminated. It might also help adding a comment to that catch like:
```
// Ignored: Typically, when UserLibrary::stopWatching is invoked, an interruptionException is triggered to stop
// this watch service 
```
that would also help understanding why `finalize` is needed.

kit/src/main/java/com/oracle/javafx/scenebuilder/kit/library/user/LibraryFolderWatcher.java
```diff
@@ -161,98 +161,110 @@ private void runDiscovery() throws InterruptedException {
...
-
-            WatchService watchService = null;
-            while (watchService == null) {
-                try {
-                    watchService = folder.getFileSystem().newWatchService();
-                } catch(IOException x) {
-                    System.out.println("FileSystem.newWatchService() failed"); //NOI18N
-                    System.out.println("Sleeping..."); //NOI18N
-                    Thread.sleep(1000 /* ms */);
+        WatchService watchService = null;
```
jperedadnr, line 164: Update copyright year in license header

kit/src/main/java/com/oracle/javafx/scenebuilder/kit/library/user/LibraryFolderWatcher.java
mimmoz81: Well, there is already a catch at line 102 that catches both runDiscovery() and runWatching().
In my opinion it is better to leave both methods with the throws InterruptedException as they are and have a single catch clause that handles both. Infact an interrupt could happen just before runWatching() during runDiscovery().
Having a single catch block is better even if in the future there will be another runSomething() following runWatching()..
It is ok to better clarify with a comment atone 102 and, I add, to log at info level the end of the service/thread.

jperedadnr: That makes sense. Just maybe include the comment I suggested to the finally clause? No need to log at info level though.

mimmoz81: Ok I'll add the comment. I'm out for vacation these days. I will be back next monday.

kit/src/main/java/com/oracle/javafx/scenebuilder/kit/library/user/LibraryFolderWatcher.java
mimmoz81: Ok will do that

mimmoz81:
@jperedadnr comments added and year updated

jperedadnr (approved):
Looks good

PR #396, opened 2021-08-29
feat: Version specific store of app data, and preferences 

### LINKED ISSUE #346: Cannot open multiple instances of Scene Builder (e.g. to compare SB11 vs. SB15 or SB16)
Starting multiple instances of SceneBuilder does not work.

### Expected Behavior

* running `SceneBuilder.exe` (from link or from CLI) shall open a new Scene Builder window
* repeated execution of `SceneBuilder.exe` should open a new window backed by an independen new instance
* Logfile shoud indicate issue
* As JavaFX shows warnings when using different FXML versions, for some scenarios it can be useful to run multiple instances of SceneBuilder.
* If it is intentional to only have one instance, question: Would it be helpful to allow multiple instances for different Scene Builder versions?

### Current Behavior

* Opening multiple windows by creating a new view from template or opening a FXML file works.
* Starting a single instance using `SceneBuilder.exe` works.
* Starting a second instance of `SceneBuilder` (regardless if same version or different version) does not work
* Logfile is empty

### Steps to Reproduce

* Open new instance of Scene Builder 16
* Attempt to open second instance of Scene Builder 16/15/11
* Attempt to run SceneBuilder using Maven or from within Eclipse
* In all attempts, I was not able to obtain any useful information from Scene Builder logfile.

Running tests entirely from IDE helped me as follows:

* Start first Scene Builder instance from within Eclipse using a Run Configuration.
* Attempt to start second instance ... will fail with a stack trace:

```
Apr. 01, 2021 6:18:37 NACHM. com.oracle.javafx.scenebuilder.app.SceneBuilderApp$SceneBuilderUncaughtExceptionHandler uncaughtException
SEVERE: An exception was thrown:
java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:64)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:70)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:248)
	at java.base/java.util.Objects.checkIndex(Objects.java:372)
	at java.base/java.util.ArrayList.get(ArrayList.java:458)
	at com.oracle.javafx.scenebuilder.app.SceneBuilderApp.handleOpenFilesAction(SceneBuilderApp.java:506)
	at com.oracle.javafx.scenebuilder.app.AppPlatform$MessageBoxDelegate.lambda$0(AppPlatform.java:186)
	at javafx.graphics/com.sun.javafx.application.PlatformImpl.lambda$runLater$10(PlatformImpl.java:447)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at javafx.graphics/com.sun.javafx.application.PlatformImpl.lambda$runLater$11(PlatformImpl.java:446)
	at javafx.graphics/com.sun.glass.ui.InvokeLaterDispatcher$Future.run(InvokeLaterDispatcher.java:96)
	at javafx.graphics/com.sun.glass.ui.win.WinApplication._runLoop(Native Method)
	at javafx.graphics/com.sun.glass.ui.win.WinApplication.lambda$runLoop$3(WinApplication.java:174)
	at java.base/java.lang.Thread.run(Thread.java:834)
```

### Environment

* Windows 10 Pro
* Version	20H2
* Build 19042.867
* Windows Feature Experience Pack 120.2212.551.0

### Screenshots

SceneBuilder2ndInstanceStackTraceAtLaunch

### LINKED ISSUE #369: scenebuilder overwrites preferences when installed multiple versions and change preferences in one of them
When i install two versions of scenebuilder ex: scenebulider 16 and scenebuilder 8 and change preferences in one of the instance  it writes to both versions of scenebuilder and when i install plugins from library manager they become avaidable in both versions of scenebuilder that is not desired

### LINKED ISSUE #395: SceneBuilder is opening slow when you have Intelij open
When starting scenebuilder from intelij idea ,Right CLick on fxml file in the projects view , then choose open in scene builder 
it open really slow , and some times it does not open scenebuilder at all , you have to close intelij idea then scenebuilder open imidately

Oliver-Loeffler:
Application data, User Library and Preferences are now stored specific to each SceneBuilder version. Thus multiple versions can run in parallel.

Until now, all Scene Builder versions shared the same settings tree in file system (application data) which prevented parallel execution of Scene Builder versions. It was not possible, to compare e.g. SB11 with SB16 or SB8 side by side. This works now.

Also, user library was shared amongst all SB versions which in some cases lead to trouble as e.g. when there were multiple versions of the same control for different JDKs.

### Issue

Fixes #369
Fixes #395 (partially)
Fixes #346

### Progress


Oliver-Loeffler:
There is one question - its about the "Manage Repositories" function. The configuration behind (its all in Preferences) is basically shareable between all versions. So this one could be designed as ashared element. Otherwise users probably wil complain, that they have to import their repository settings manually between different SB versions.

What do you think? This could be a different PR.

mimmoz81:
As I said in #369 I think SB should store all settings per workspace location

Oliver-Loeffler:
Okay, with the current status each SceneBuilder has its own repositories setup - which is not a workspace yet but each version has its own config. As a next step, in order to enable workspaces later, another PR could introduce the capability to save settings to files and read them from files.

@mimmoz81 
For a possible workspace configuration, would you prefer a single file which holds all details or an sub directory such as `.scenebuilder.workspace` which holds settings in individual files?

mimmoz81:
@Oliver-Loeffler I was thinking about that too..
A single json file should be enough to represent the entire ws settings and all the needed stuffs.
A folder though is more flexible..

mimmoz81:
We should talk about that in a specific issue I think

Oliver-Loeffler:
Issue #14 is probably the better space. Just found this one.

abhinayagarwal:
How do we make sure if an existing application is being updated, the newer version has all the settings/preferences from the older version?

Oliver-Loeffler:
At least the following approaches are thinkable:

## Approach 1
* on first start of new SB version, scan for settings of previous versions
* show a list of the versions found
* ask user to import settings from one of the previous versions or to start from scratch

## Approach 2
* on first start of new SB version, scan for settings of previous versions
* silently import settings of the most recent installed version

What do you think?

abhinayagarwal:
I am more inclined towards Approach 2

johanvos:
I like the NetBeans approach here: when installing a new version, it says it detected an old version, and recommends copying the data from that old version. 
So that would be a manual action, but the default would be to copy.
Hence, I have a slight preference for option 1, but I can live with both.

Oliver-Loeffler:
I will have more time to dig into this again starting next week. Personally, I do prefer approach 1 over approach 2, I appreciate being asked over being forced to do things in a certain way. For that, let's see, I'll prepare a prototype for each approach so we can test this.

Oliver-Loeffler:
It took me a moment to get back into it but now I have a plan to create the prototype. I'm going to use the existing PR for that. Also it can be helpful to document where and how SceneBuilder stores its settings. As of now its only a a markdown snippet in `docs/snippets`.

Oliver-Loeffler:
There is now the first prototype. It detects settings of older SB versions (well, until now its always SB_2.0) and asks one single time at first start if an import is desired. The user decision is memorized.

It works like that:
* There is a type `AppVersion` which knows how to handle version numbers such as `1.0` or `1.1.1`, a version such as `1.1.0` is considered to be more recent than `1.1` or `1.0`.
* Then there is also a type `VersionedNode` which combines an AppVersion with a Preferences node.
* The PreferencesController now has the ability to provide a PreferencesImporter.
* The import process takes place in `SceneBuilderApp.handleLaunch()` method, there the method `PreferencesImporter .askForActionAndRun()` is called where the process takes place.

The `PreferencesController` parametrizes the `PreferencesImporter` accordingly, so that the necessary inits after import take place, which are e.g. reload Maven settings or Repository settings. Otherwise a restart of SB would be needed after import.

As of now, all settings which live in Java Preferences world are imported. I have not yet studied if it is necessary to import the MessageBox. Another open point is, importing the user JAR file repository. This is not implemented yet but my proposal here is, also to ask the user (e.g. by adding a checkbox to the import question alert).

@abhinayagarwal What do you think?

Oliver-Loeffler:
The current draft version also imports the User Library folder. But this is clearly a draft, I am not happy yet with the OS specific handling of directories - especially with regard to testability. So, there will be another commit as follow up. But it would be great if you could look into this.

May be you can help me @abhinayagarwal:
The library directory can be big, so importing a user library can be time consuming. Would it be desirable to indicate progress for this cases?

abhinayagarwal:
Feedback is always important to the user. If an application doesn't respond for a few seconds, it can lead to bad UX. Therefore, if you think that import will take more than a second, I preference would be to add a progress indicator somewhere.

Oliver-Loeffler:
Okay, so let me prepare a dialog with progress bar for the import (only for the part when libraries are imported), the preferences part works really quick.

Oliver-Loeffler:
Hello,

first of all my apologies, this one turned out getting larger than expected - the import was partially kind of tricky.
But now this one has reached a state of work, where I'd like to wait for your feedback before continuing.

After playing a while, I'm with @abhinayagarwal (approach 2) to do all the imports silently and never ask.

Goals of this PR:
* enable co-existance of different Scene Builder versions with their individual settings
* each version should have its own settings
* newer versions can import existing settings and user library contents (added during discussion)
* all settings directories are now version specific (so its no longer `AppData\Roaming\Scene Builder` but instead `AppData\Roaming\Scene Builder 17.0.0`. 
* newer versions do not touch any settings of older versions, both can coexist
* The user "dot" directory `~/.scenebuilder` remains the same but the log file is now version specific (i.e. `scenebuilder-17.0.0.log.0`)

It works as follows:
* Scene Builder launches
* The `PreferencesImporter` searches for preferences of older versions.
* It will ask to import - if answered yes it will do so
* The import decision is shared with `UserLibraryImporter` and it will start a JavaFX task to import user library in background.
* 2 preference values are stored to prevent future imports (one for preferences import, one for user library import).
* If one wants to enforce an import, run Scene Builder with `-DforceImport=true` option.

This PR consists of 3 larger changes:

To achieve this goal and to get most components of this tested, following work has been performed:

## 1 - version specific settings
* Operating system detection moved into an enum type `OperatingSystem`
* All the platform specific handling of settings directories and log files has been moved into `PlatformSpecificDirectories` class.
* This all is used inside `AppPlatform` and now it allows to test almost everything.
* The message box directory is not part of `AppPlatformDirectories` interface as the MB directory resides inside the application data dir.
* Where which settings are stored and what each folder means is documented in [AppConfiguration.md](https://github.com/Oliver-Loeffler/scenebuilder/blob/issue-346/docs/snippets/AppConfiguration.md)

AppPlatformDirectories

## 2 - Import of application specific preferences
* There is now a class `PreferencesImporter` which can import the full application preferences tree of a previous verstion.
* To distinguish versions and to provide an order (what is the previous one...) a record type `AppVersion` was introduced.
* When multiple preferences trees exist, for each preferences tree a version can be assigned (e.g. `SB_2.0` or `SB_17.0.1` or `SB_18.0.0-SNAPSHOT`, hence a record `VersionedPreferences` was introduced.
* The `VersionedPreferencesFinder` class then walks the Scene Builder preferences tree and searches for possibly existing older version settings.
* The app `PreferencesController` now allows to get an instance of the `PreferencesImporter` and there one can do the job like:

```java
    PreferencesImporter prefsImporter = PreferencesController.getSingleton().getImporter();
    boolean doImport = prefsImporter.askForActionAndRun();  
    // the boolean is currently only returned in order to memorize user decision - we can remove this later
```

* The `PreferencesImporter` also ensures that the `PreferencesController` reads all the settings again after completing the import. Otherwise a restart of Scene Builder would be needed.
* If an import is needed is also stored in preferences. 
* For test purposes the import can be enforced by starting Scene Builder with `mvn javafx:run -f app -DforceImport=true`
* The alert asking for action has not received much care yet as I think its better to review the existing concept first.

PreferencesImport

## 3 - Import of user library
* This task is handled by `com.oracle.javafx.scenebuilder.app.library.user.UserLibraryImporter`
* This will only be executed if previous Scene Builder preferences exist - so if only the library directory exists it won't work.
* As runtime was a concert here, I've selected a JavaFX task to do the job, thus all work is performed in background and the user basically won't notice it during startup

```java
    // The import will start as soon as the user library is initialized.
    // Import status, if triggered, if failed or completed, will be written to log.
    if (doImport) {
        Platform.runLater(UserLibraryImporter.createImportTask());
    }
```

UserLibImporter1

UserLibraryImporter2

app/src/main/java/com/oracle/javafx/scenebuilder/app/preferences/PreferencesImporter.java
```diff
@@ -180,14 +176,24 @@ public void tryImportingPreviousVersionSettings() {
...
+        if (forcedImport) {
+            logger.log(Level.FINE, "detected -DforceImport=true");
+        }
+        return previousVersionFound && (forcedImport || askForImport());
     }
 
-    private VersionedPreferences previousVersionPrefsFound(VersionedPreferences versionedpreferences) {
-        logger.log(Level.FINE, "{0}", versionedpreferences);
-        return versionedpreferences;
+    private boolean importForced() {
```
eugener, line 193: IMO, a better name for this method would be `isImportForced`

Oliver-Loeffler: Before renaming it, as I've added this mainly for debugging.
is this option something desirable to keep in the code?

eugener: To me, it makes sense assuming functionality of this PR is accepted.

Oliver-Loeffler: Okay, good to know. Actually, is it possible to pass properties to the execution when staring Scene Builder using Maven? I tried to run `mvn javafx:run -f app -DforceImport=true` and unfortunately this does not work.

eugener: If I'm not mistaken, you have to pass a property for javafx maven plugin. 
Since command line args for the plugin are under `<commandlineArgs>` tag, your argument should look somewhat like: `-DcommandlineArgs="-DforceImport=true"`. 
I haven't tested this, but this is how Maven passed arguments to plugins.

Oliver-Loeffler: Thank you - I tried this but unfortunately this did not work. However, running the Jar manually or running from within IDE works.

Oliver-Loeffler:
@AlmasB Hi Almas, I'll make this one fit to properly merge with the current main line. 
Would it possibly be helpful If I start a new PR with the desired change so that we can review the whole change set?
The previous discussion has lead to changes I had not anticipated upfront. What do you think?

AlmasB:
@Oliver-Loeffler 

I've only skimmed through this, but I was wondering if you can split this into multiple smaller PRs: one for tests, one for production code change, one for refactor, etc. That way it's easier to review and manage scope. See if that is possible?

Oliver-Loeffler:
Hello @AlmasB, exactly as you proposed, I'll rework this one and re-issue smaller PRs.

Issue #327, opened 2021-03-07
Dynamic scene size breaks scene builder


delvh:
If a project uses dynamic scene sizes by using the current screen, i.e.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<?import javafx.scene.layout.GridPane?>
<?import javafx.stage.Screen?>
<GridPane maxHeight="-Infinity" maxWidth="-Infinity"
	minHeight="400.0" minWidth="500.0"
	prefWidth="${screen.visualBounds.width}"
	prefHeight="${screen.visualBounds.height}"
	xmlns="http://javafx.com/javafx/11.0.1"
	xmlns:fx="http://javafx.com/fxml/1">
	<fx:define>
		<Screen fx:factory="getPrimary" fx:id="screen" />
	</fx:define>
</GridPane>
```
scene builder will refuse to load the fxml file, or even crash when called from for example the `e(fx)clipse` plugin for Eclipse.

JavaFX however will accept the fxml file and run as intended.

Additionally, on some platforms it appears that after such a file has been encountered, scene builder is unable to terminate when closed. Only after manually killing SceneBuilder (i.e. with SIGKILL), can you reopen it. This is bad because it seems that only one instance of SceneBuilder can be open, resulting in effectively a dead program.

### Expected Behavior
Scene builder should be able to load the file, at best with the size as intended.

### Current Behavior
Scene builder does not load the file when opened using internal `Open` button, or does not start at all when called externally, i.e. through the `e(fx)clipse` plugin for Eclipse.

### Steps to Reproduce
Extract the code given above into an fxml file, replace contro``
Try to open the code mentioned above in scene builder: Scene builder will not open the given code file

### Your Environment
Has been observed on many Linux platforms, including `Ubuntu 18.04`, `Kubuntu 20.04`, `Gentoo stable`, `Arch 1.4`.

### Screenshots
Not relevant, as simply **nothing** happens.

Oliver-Loeffler:
This is related to #402, with PR #403 an error message will be shown to the user.
There is the following stack trace then:

```
java.lang.IllegalStateException: Bug in FXOMRefresher: FXML dumped in /var/folders/kd/22d474z97yx97rgj2ttf_c3r0000gn/T/DTL-5996-6816594968554613127.fxml
	at com.oracle.javafx.scenebuilder.kit.fxom.FXOMRefresher.refresh(FXOMRefresher.java:96)
[...]

Issue #253, opened 2019-08-19
SB won't load a file that contains <userData>


bendem:
I have the file 

```xml
<?xml version="1.0" encoding="UTF-8"?>

<?import javafx.geometry.Insets?>
<?import javafx.scene.control.Button?>
<?import javafx.scene.control.TableColumn?>
<?import javafx.scene.control.TableView?>
<?import javafx.scene.layout.HBox?>
<?import javafx.scene.layout.VBox?>

<VBox prefWidth="940.0" xmlns="http://javafx.com/javafx/11.0.1" xmlns:fx="http://javafx.com/fxml/1" fx:controller="io.a2l.extenso.kitdev.controllers.MethodFilesController">
    <userData>
        <fx:reference source="controller"/>
    </userData>
   <children>
      <HBox spacing="7.0">
         <children>
            <Button fx:id="importButton" mnemonicParsing="false" onAction="#onImport" text="Import methods" />
            <Button fx:id="exportButton" mnemonicParsing="false" onAction="#onExport" text="Export selection" />
            <Button fx:id="analysisButton" mnemonicParsing="false" onAction="#onAnalysis" text="Use method for analysis" />
            <Button fx:id="deleteButton" mnemonicParsing="false" onAction="#onDelete" text="Delete selected methods" />
         </children>
         <opaqueInsets>
            <Insets />
         </opaqueInsets>
         <padding>
            <Insets bottom="7.0" left="7.0" right="7.0" top="7.0" />
         </padding>
      </HBox>
      <TableView fx:id="tableView" minWidth="700.0" VBox.vgrow="ALWAYS">
        <columns>
          <TableColumn fx:id="kitId" editable="false" prefWidth="178.0" text="Kit ID" />
          <TableColumn fx:id="methodId" editable="false" prefWidth="164.0" text="Method ID" />
          <TableColumn fx:id="revision" editable="false" prefWidth="148.0" text="Revision" />
        </columns>
      </TableView>
   </children>
</VBox>
```

I have to remove the `<userData>` tag, before opening the file, otherwise I get

```
Aug 19, 2019 5:14:43 PM com.oracle.javafx.scenebuilder.app.SceneBuilderApp$SceneBuilderUncaughtExceptionHandler uncaughtException
SEVERE: An exception was thrown:
java.lang.NullPointerException
        at com.oracle.javafx.scenebuilder.kit.fxom.FXOMLoader.load(FXOMLoader.java:98)
[...]

PR #403, opened 2021-08-29
When loading of FXML file fails, in some cases the error is not shown to the user

### LINKED ISSUE #402: For some cases on errors during FXML Loading no error dialog is presented
As described in issue #253, some FXML files failed to load.
The case described in #253 fails in a way, that the error is not communicated to the user.

### Expected Behavior

Each error which prevents FXMLFiles from being loaded shall be presented to the user.

### Current Behavior

SceneBuilder remains silent and does not load the FXML.
The error is logged.

### Steps to Reproduce

Save the attached code as `broken.fxml` and attempt to load.
```fxml
<?xml version="1.0" encoding="UTF-8"?>

<?import javafx.geometry.Insets?>
<?import javafx.scene.control.Button?>
<?import javafx.scene.control.TableColumn?>
<?import javafx.scene.control.TableView?>
<?import javafx.scene.layout.HBox?>
<?import javafx.scene.layout.VBox?>

<VBox prefWidth="940.0" xmlns="http://javafx.com/javafx/11.0.1" xmlns:fx="http://javafx.com/fxml/1" fx:controller="some.unknown.controller">
    <userData>
        <fx:reference source="controller"/>
    </userData>
   <children>
      <HBox spacing="7.0">
         <children>
            <Button fx:id="importButton" mnemonicParsing="false" onAction="#onImport" text="Import methods" />
            <Button fx:id="exportButton" mnemonicParsing="false" onAction="#onExport" text="Export selection" />
            <Button fx:id="analysisButton" mnemonicParsing="false" onAction="#onAnalysis" text="Use method for analysis" />
            <Button fx:id="deleteButton" mnemonicParsing="false" onAction="#onDelete" text="Delete selected methods" />
         </children>
         <opaqueInsets>
            <Insets />
         </opaqueInsets>
         <padding>
            <Insets bottom="7.0" left="7.0" right="7.0" top="7.0" />
         </padding>
      </HBox>
      <TableView fx:id="tableView" minWidth="700.0" VBox.vgrow="ALWAYS">
        <columns>
          <TableColumn fx:id="kitId" editable="false" prefWidth="178.0" text="Kit ID" />
          <TableColumn fx:id="methodId" editable="false" prefWidth="164.0" text="Method ID" />
          <TableColumn fx:id="revision" editable="false" prefWidth="148.0" text="Revision" />
        </columns>
      </TableView>
   </children>
</VBox>
```

### Your Environment
* OS:  Windows 10 Pro
* Version: 21H1
* Build: 19043.1165
* SceneBuilder 16
* Java 16 / JavaFX 16

Oliver-Loeffler:
When loading of FXML file fails, in some cases the error is not shown to the user.
This is actually the case for #253. 

The error detection in `SceneBuilderApp` only worked for cases of `IOExceptions`.
This case is caused by a `NullPointerException` coming from `FXMLLoader` inside `FXOMLoader`.

I cannot solve the issue #253 now but consider this behavior, that no error dialog is presented, as a bug.

### Issue

Fixes #402
Related #327
Related #253
Related #54
(there are more very similar issues)

### Progress


kit/src/test/resources/com/oracle/javafx/scenebuilder/kit/fxom/BrokenByUserData.fxml
```diff
@@ -0,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<?import javafx.geometry.Insets?>
+<?import javafx.scene.control.Button?>
+<?import javafx.scene.control.TableColumn?>
+<?import javafx.scene.control.TableView?>
+<?import javafx.scene.layout.HBox?>
+<?import javafx.scene.layout.VBox?>
+
+<VBox prefWidth="940.0" xmlns="http://javafx.com/javafx/11.0.1" xmlns:fx="http://javafx.com/fxml/1" fx:controller="io.a2l.extenso.kitdev.controllers.MethodFilesController">
```
abhinayagarwal, line 10: Can the controller package be changed to something more generic? e.g. `com.example.controller`

Oliver-Loeffler: Very good point. Thanks!

kit/src/test/java/com/oracle/javafx/scenebuilder/kit/fxom/FXOMDocumentTest.java
```diff
@@ -0,0 +1,51 @@
...
+public class FXOMDocumentTest {
+    @Test
+    public void that_IOexception_is_thrown_in_case_FXMLLoader_error() throws Exception {
+        URL resource = getClass().getResource("BrokenByUserData.fxml");
+        String fxmlText = FXOMDocument.readContentFromURL(resource);
+        Throwable t = assertThrows(IOException.class, () -> new FXOMDocument(fxmlText, resource, null, null));
+        String message = t.getMessage();
+        assertTrue(message.startsWith("javafx.fxml.LoadException:"));
+    }
+}
```
abhinayagarwal, line 51: Can we add a few more test cases to assert both known and unknown causes when loading a FXML? May be also assert a case where `XMLStreamException` is thrown.

Oliver-Loeffler: Yes, of course. This is a good point.

abhinayagarwal: Hi @Oliver-Loeffler,

Are you going to add some more test cases to move this PR forward?

Oliver-Loeffler: Yes, there will be an update with more tests.

Oliver-Loeffler: Hi @abhinayagarwal 
There is a significant overlap with PR #405 hence I am waiting for your feedback regarding #405 before continuing here. When #405 becomes accepted, I'll rebase towards 405 and continue from there.

Oliver-Loeffler: There is also now a case for a `SAXParseException` which is wrapped as cause in a `IOException`. This is the case when it is attempted to load and incomplete or broken XML structure.

Oliver-Loeffler:
@abhinayagarwal Do you have any idea on how to really ensure that the test case `FXOMDocumentTest.that_missing_imports_during_defines_resolution_cause_exception()` is safely executed on Windows?
The expectation there is an `IllegalStateException` caused by `FXOMRefresher`. It works fine locally with JavaFX 17 on Temurin 11 but continuously fails in GH actions.

kit/src/test/java/com/oracle/javafx/scenebuilder/kit/fxom/FXOMDocumentTest.java
Oliver-Loeffler: @abhinayagarwal  Eventually it worked and now I have a case where the XMLStreamException is provoked and handled.

Oliver-Loeffler:
@AlmasB If you like we can start with this one. Would be glad if you'd have an eye on this one. We have many older (some newer) issues related to FXML/FXOM loading. This PR is reviewed by @abhinayagarwal . Goal here is at first, that Scene Builder no longer silently ignores exceptions. There is actually a mechanism which displays the Exception contents to the user but this was not triggered in past. So users complained about FXMLs which failed to load.

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
```diff
@@ -668,7 +668,7 @@ private void performOpenFiles(List<File> fxmlFiles,
                     hostWindow.loadFromFile(fxmlFile);
                     hostWindow.openWindow();
                 }
-            } catch (IOException xx) {
+            } catch (RuntimeException | IOException xx) {
```
AlmasB, line 671: Is there a reason not to use `Exception` superclass here, i.e. why `Runtime` and `IO` only?

kit/src/main/java/com/oracle/javafx/scenebuilder/kit/fxom/FXOMLoader.java
```diff
@@ -95,14 +103,31 @@ public void load(String fxmlText) throws java.io.IOException {
...
+    private void handleFxmlLoadingError(Exception x) throws IOException {
+        if (x.getCause() != null) {
+            handleKnownCauses(x);
+        } else {
+            handleUnknownAndMissingCauses(x);
+        }
+    }
+
+    private void handleUnknownAndMissingCauses(Exception x) throws IOException {
+        throw new IOException(x);
```
AlmasB, line 119: I assume throwing this is fine as it will be caught by the SB general exception handler?

AlmasB (approved):
Looks good, please can you check the two questions.

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
Oliver-Loeffler: Very good point, yes, you are absolutely right. This is possibly a leftover from searching what was happening. Catching any kind of exception is a better solution here.

kit/src/main/java/com/oracle/javafx/scenebuilder/kit/fxom/FXOMLoader.java
Oliver-Loeffler: I do not actually know why there was a separation between the path where causes are shown and where just an IOException is raised. The branching was already there. Technically, I think, that the branching is most likely not needed and we just could use the exception handler here to show the error. 

IOException and Runtime exceptions will eventually cause an IllegalStateException in FXOMRefresher and the FXML causing the problem will be dumped. There is no proper error handling yet which becomes visible in GUI.

As of now, I would not change this as this exceeds the scope of the PR and I cannot oversee consequences yet. For further changes here, I'd like to add more tests first.

AlmasB: ok

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
AlmasB: If you wish to update this, happy to wait. Otherwise, it looks good to go.

Oliver-Loeffler (commented):
ok

PR #406, opened 2021-09-07
Containment/fix for issue #373

### LINKED ISSUE #373: [Linux] Main window size changes when the update popup shows up
The issue was forked out of #352

Scene Builder window is maximed initially, but gets shrunk when the update dialog appears.

### Expected Behavior
main window size is not changed.

### Current Behavior
 main window size is reduced.

### Steps to Reproduce

- Start SceneBuilder with mvn javafx:run -f app.
- Close the window with the recent FXML files opened (title "Gluon Scene Builder").
- Error popup should appear. Size of the main window is changed.

### Your Environment
Linux build from master

### Screenshots
scenebuilder-update-check

Oliver-Loeffler:
Before a update check dialog is shown in `SceneBuilderApp`, the current location and size of the document window is memorized.
Using the `setOnShowing` hook the window size is restored in the case it has been altered by calling `SBAlert.showAndWait`.

This PR is experimental as I think, there is something else on JavaFX level not working properly. For this a separate proof-of-concept project is in works.

### Issue

Fixes #373

### Progress

### DIFF
--- a/app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
+++ b/app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
@@ -792,9 +792,7 @@ public class SceneBuilderApp extends Application implements AppPlatform.AppNotif
         }
     }
 
-    private enum ACTION {START, STOP}
-
-    ;
+    private enum ACTION {START, STOP};
 
     private void logTimestamp(ACTION type) {
         switch (type) {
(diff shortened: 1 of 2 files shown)


abhinayagarwal:
This feels more like a hack than a solution. Copied my findings from  internal issue tracker to #373

Oliver-Loeffler:
Thats why I've called it acontainment.

abhinayagarwal:
Would be really nice if we could fix the core issue..

Oliver-Loeffler:
Yes, this is the preferred way. So we can close this PR.

abhinayagarwal:
Closing PR. It would be nice to get the root cause of the issue.

Oliver-Loeffler:
@abhinayagarwal Do you know if there is already a JBS entry for the issue we observed here?

PR #410, opened 2021-09-13
release-drafter: categorize changes

abhinayagarwal:
Release drafter is used to automatically create release notes for Scene Builder. Since we have quite a few changes for 17, it is a good idea to categorize them.

All PRs are now labelled as either 'enhancement' or 'bug' . We use these labels against these PRs to group them into 'Features' and 'Bug Fixes' in the release notes.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,5 +1,12 @@
 exclude-labels:
   - 'housekeeping'
+categories:
+  - title: '🚀 Features'
+    labels:
+      - 'enhancement'
+  - title: '🐛 Bug Fixes'
+    labels:
+      - 'bug'
 template: |
   ## What’s Changed
   $CHANGES


erwin1 (approved):

Oliver-Loeffler:
Are the release notes reviewed and possibly edited before release?

abhinayagarwal:
Yes, they are. Do you have any feedback on the [current draft](https://github.com/gluonhq/scenebuilder/releases)?

Oliver-Loeffler:
I am just curious about the release process in general. Reading through the commits I found, that not all contain references to issues. Also its quite cool to see, that you generate the release notes automatically, which eventually shows that there are some quality requirements to commit messages - which, is good to know.

What I have noticed up to that point, we should describe the time-boundary condition until when PRs have to get merged to become part of a release. There is definitely a timeframe around a release, where PRs may not get added anymore.
A good place for such a note could be the `CONTRIBUTING.md` once it exists.

abhinayagarwal:
> What I have also noticed is, that we should describe the time-frame before a release until when PRs have to be merged to become part of a release. When this time-frame is passed, some PRs may not become part of the next release. This is something to communicate, e.g. in the future CONTRIBUTING.md.

This is exactly the reason why I have created milestones. We can mark the PRs which can make it to a definitive release, or if for some reasons they fail to make it into a release, we can retarget them to the next. We currently do not have a timeline for releases, each release is made within a week or 2 of  the Java release. Although, I agree adding a timeline for freezing commits to a release is a good idea.

> Unfortunately I cannot see a SB17 release notes draft yet.

Oops! Check this screenshot.

image

Oliver-Loeffler:
Ah, its the titles of the PRs which are used. Very neat so its worth to think about a good title for the PR.
Having the milestones is a good idea, this gives a good orientation. Thanks for sharing this.

PR #427, opened 2022-01-05
fix: Added workaround for JDK-8205915. Will fix issues #221, #263, #306 and 556.

### LINKED ISSUE #221: Pasted values are duplicated in section Properties -> Style
I'm using:
```
Product Version
JavaFX Scene Builder 10.0.0

Build Information
Version 10.0.0
Date: 2018-06-05
Java Version: 10.0.1+10, Oracle Corporation
....
....
Operating System
Mac OS X, x86_64, 10.13.6

```

whenever I copy a value and paste it to the **Properties -> Style** I see the value is duplicated in one field: `#2b4496#2b4496`

For example, I want to use color `#2b4496` for style `-fx-background-color` -> I copy the value `#2b4496` somewhere and paste it as value in the scene builder and get the value `#2b4496#2b4496`. It always happens when I use hot keys for copy/paste in my Mac OS Command+C/Command+V, but it works just fine when I use mouse right click -> copy -> paste.

### LINKED ISSUE #263: Pasting Selected Text in Code Inspector cause duplicate pasting
JDK:11,12
JavaFX versions: 11,12,13
Operating System Tested:Mac OSX Mojave 10.14.6

How to reproduce:
1. Click on any TextInputControl in UI designer
2.Expand Code Section in Inspector
3. Type in value 'Hello' into any field (OnAction, On Drag Detected and so on)
4. Select the text and paste into any other empty field
5. The test pasted is 'HelloHello'

You can fix the problem by commenting out the paste invocation (not a proposed solution BTW!) in com.oracle.javafx.scenebuilder.app.DocumentWindowController::performPaste.

    private void performPaste() {
        final Node focusOwner = this.getScene().getFocusOwner();
        // If there is FXML in the clipboard, we paste the FXML whatever the focus owner is
        if (getEditorController().canPerformEditAction(EditAction.PASTE)) {
            this.getEditorController().performEditAction(EditAction.PASTE);
            // Give focus to content panel
            contentPanelController.getGlassLayer().requestFocus();
        } else {
            assert isTextInputControlEditing(focusOwner);
            final TextInputControl tic = getTextInputControl(focusOwner);
            // tic.paste();  <====  HACK - PREVENTING PASTE FIXES THE ISSUE!
        }
    }

Somewhere inside  **TextInputControl::paste** it is duplicating the pasted text.
The issue has been around for a while, and has been bugging me, so I will toss my hat into the ring and attempt a fix. 

Wanted to get some feedback first, before attempting my first PR.

1. I have attached a FXML file (with .txt extension) 
2. a screen capture video can be found here - [https://drive.google.com/open?id=1G02417zomzeuwpPckcvVuPUrUY3IS30a](https://drive.google.com/open?id=1G02417zomzeuwpPckcvVuPUrUY3IS30a)

[pastebug_fxml.txt](https://github.com/gluonhq/scenebuilder/files/3981866/pastebug_fxml.txt)

### LINKED ISSUE #306: Values pasted into text fields are duplicated
If you paste a value into the, e.g., "On Drag Detected" field it will get duplicated. I.e., if you copy something like "doIt" from somewhere and past it into the said field, you will end up with "doItdoIt" in that field. I tested this on a macBook Pro with SceneBuilder 15 RC1. (Actually this is not a new bug because it can be observed in older versions too.)

### LINKED ISSUE #556: "Double paste" when pasting in text field
When you paste something (e.g. a fx:id) in a text field it is pasted twice.

### Expected Behavior
Example: I copy a text e.g. "okButton" and paste it into the fx:id field I expect that in the text field now should stand "okButton".

### Current Behavior
Let's use the same example with the okButton. When I paste "okButton" into the text field it does not say "okButton" but "okButtonokButton".

### Steps to Reproduce
Copy a text and paste it into a text field.

### Your Environment
I'm currently using a MacBook Pro with macOS 10.15.7 (Catalina) and Scene Builder 18.

### Screenshots
Bildschirmfoto 2022-04-30 um 12 22 17
Pasted "okButton"

Oliver-Loeffler:
To avoid the issue filed in JDK-8205915, for MacOS the call to `TextInputControl.paste()` method is replaced with a class `com.oracle.javafx.scenebuilder.app.util.TextInputControlPaster`. This utility class prevents the controls paste method if desired and directly writes the Clipboard contents into the correct position of the TextInputControl. 

This does not fix the root cause for JDK-8205915 but ensures that the paste functionaliy can be used in Scene Builder on MacOS. Once JDK-8205915 is solved, this workaround shall be removed.

### Issue
When pasting clipboard contents into TextInputControl elements, then the clipboard contents is always pasted twice on MacOS. There is a bug ticket for JavaFX in JBS, which is JDK-8205915.

As this behavior is very annoying on MacOS, this PR might provide a temporary workaround until JDK-8205915 is fixed.

Fixes #221
Fixes #263
Fixes #306
Fixes #556

### Progress


app/src/main/java/com/oracle/javafx/scenebuilder/app/util/TextInputControlPaster.java
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * This class provides a workaround for the issue in JavaFX
+     * which prevents Scene Builder from pasting text values properly
+     * in TextInputControl elements on MacOS. If attempted, clip board
+     * contents is always pasted twice.
+     * 
+     * @see https://bugs.openjdk.java.net/browse/JDK-8205915
+     */
+    public TextInputControlPaster(TextInputControl tic) {
+        this(tic, ()->Clipboard.getSystemClipboard().toString());
```
jperedadnr, line 55: add spaces before and after lambda

app/src/test/java/com/oracle/javafx/scenebuilder/app/JfxInitializer.java
```diff
@@ -0,0 +1,64 @@
+/*
```
jperedadnr, line 1: Too bad we need to duplicate this class in `app`. Publishing test-kit for the sake of adding it as a dependency might be overkilling just for this?

app/src/test/java/com/oracle/javafx/scenebuilder/app/util/TextInputControlPasterTest.java
```diff
@@ -0,0 +1,106 @@
...
+        assertEquals("1234567890", textField.getText());
+        
+        classUnderTest.pasteWorkaroundOnMacOs(null);
+        assertEquals("1234567890", textField.getText());
+    }
+    
+    @Test
+    public void that_workaround_is_applied_conditionally() throws Exception {
+        TextField textField = new TextField();
+        classUnderTest = new TextInputControlPaster(textField, ()->"Text2Paste");
```
jperedadnr, line 81: add spaces to lambda

app/src/main/java/com/oracle/javafx/scenebuilder/app/util/TextInputControlPaster.java
```diff
@@ -0,0 +1,98 @@
...
+    protected void replaceSelection(String text) {
+        IndexRange indexRange = tic.getSelection();
+        insertText(text, indexRange.getStart(), indexRange.getEnd());
+    }
+    
+    private void insertText(String text, int start, int end) {
+        String ticText = tic.getText();
+        String prefix = ticText.substring(0, start);
+        String suffix = ticText.substring(end, ticText.length());
+        tic.setText(prefix+text+suffix);
```
jperedadnr, line 96: Add spaces before and after `+`

app/src/main/java/com/oracle/javafx/scenebuilder/app/util/TextInputControlPaster.java
```diff
@@ -0,0 +1,98 @@
...
+
+    protected void replaceSelection(String text) {
+        IndexRange indexRange = tic.getSelection();
+        insertText(text, indexRange.getStart(), indexRange.getEnd());
+    }
+    
+    private void insertText(String text, int start, int end) {
+        String ticText = tic.getText();
+        String prefix = ticText.substring(0, start);
+        String suffix = ticText.substring(end, ticText.length());
```
jperedadnr, line 95: `ticText.length()` is not required?

app/src/main/java/com/oracle/javafx/scenebuilder/app/util/TextInputControlPaster.java
jperedadnr, line 55: Testing on Mac (Monterey), if I copy the text "1234" into the clipboard, select a label node, and paste it into the `Text` field in the Inspector tab I get without this PR the expected `12341234`, but with this PR I get: 
```
1234javafx.scene.input.Clipboard@5474435e
```
Changing  `() -> Clipboard.getSystemClipboard().toString()` into `() -> ""` seems to prevent that. I wonder why you have this?

In any case, I get the caret at the beginning of the text, which is inconvenient.

app/src/main/java/com/oracle/javafx/scenebuilder/app/util/TextInputControlPaster.java
```diff
@@ -0,0 +1,98 @@
...
+        insertText(text, caret, caret);
+    }
+
+    protected void replaceSelection(String text) {
+        IndexRange indexRange = tic.getSelection();
+        insertText(text, indexRange.getStart(), indexRange.getEnd());
+    }
+    
+    private void insertText(String text, int start, int end) {
+        String ticText = tic.getText();
```
jperedadnr, line 93: When I copy into Text (Inspector) I can see that `ticText` already contains the text that I'm pasting from the clipboard, so `text` is not needed?

Oliver-Loeffler: I have to rework this, thanks for your analysis and help. There is a much simpler way to achieve a proper paste behavior. I'll rework the test and the utility class accordingly. This code here will go away. The solution is, to actually not call the `tic.paste()` method at all on MacOS. There will be an update tomorrow on this.

app/src/main/java/com/oracle/javafx/scenebuilder/app/util/TextInputControlPaster.java
Oliver-Loeffler: Thanks, I'll anticipate this in future and see, if I can place a rule in Checkstyle for this.

app/src/test/java/com/oracle/javafx/scenebuilder/app/util/TextInputControlPasterTest.java
Oliver-Loeffler: Same here, I'll see to get checkstyle on this.

app/src/main/java/com/oracle/javafx/scenebuilder/app/util/TextInputControlPaster.java
Oliver-Loeffler: The code has been removed as no longer necessary.

app/src/main/java/com/oracle/javafx/scenebuilder/app/util/TextInputControlPaster.java
Oliver-Loeffler: This code has been removed, it is not needed anymore.

app/src/main/java/com/oracle/javafx/scenebuilder/app/util/TextInputControlPaster.java
Oliver-Loeffler: Also this code has been removed. It's not needed.

app/src/main/java/com/oracle/javafx/scenebuilder/app/util/TextInputControlPaster.java
Oliver-Loeffler: The new approach is now much simpler.

app/src/test/java/com/oracle/javafx/scenebuilder/app/JfxInitializer.java
Oliver-Loeffler: Actually, as of now, I'd like to avoid adding test-kit as a dependency. If we get more tests in app, its probably worth to think about what we could reuse from test-kit.

app/src/main/java/com/oracle/javafx/scenebuilder/app/DocumentWindowController.java
```diff
@@ -1573,10 +1574,25 @@ private void performPaste() {
...
+     */
+    /**
+     * This method represents a workaround for the issue in JavaFX
+     * which prevents Scene Builder from pasting text values in TextInputControl elements
+     * on MacOS.
+     * 
+     * @see https://bugs.openjdk.java.net/browse/JDK-8205915
+     */
+    private void performTextInputControlPaste(final TextInputControl tic) {
+        boolean isNotMacOS = !EditorPlatform.IS_MAC;
```
jperedadnr, line 1592: This can probably be inlined.

app/src/main/java/com/oracle/javafx/scenebuilder/app/DocumentWindowController.java
```diff
@@ -1573,10 +1574,25 @@ private void performPaste() {
         } else {
             assert isTextInputControlEditing(focusOwner);
             final TextInputControl tic = getTextInputControl(focusOwner);
-            tic.paste();
+            performTextInputControlPaste(tic);
```
jperedadnr, line 1577: Since the fix is so simple now, and given the fact that it should be removed once the JBS issue gets fixed, I don't think we need a full class just for it?

It should be as simple as:

```
final TextInputControl tic = getTextInputControl(focusOwner);
// TODO: Once the issue https://bugs.openjdk.java.net/browse/JDK-8205915 is solved, 
// this if expression should be removed.
if (!EditorPlatform.IS_MAC) {
    tic.paste();
}
```

Oliver-Loeffler: Will this PR merge with JavaFX 18?

I decided for a separate class just for testability. But you're right, let me rework this.

Oliver-Loeffler: So, its now much simpler - we kind of need to keep track of this. As soon as [openjdk/jfx#704](https://github.com/openjdk/jfx/pull/704 ) is merged, the paste operation in Scene Builder will be broken on MacOS.

jperedadnr: Indeed, I can't say when that PR will be merged. Most certainly it won't make it to 18 (jfx18 stabilisation branch scheduled for this week), and SB18 will still need this patch. In the end, we might need some JavaFX version check to apply it or not...

Oliver-Loeffler: Okay, my proposal here is, to introduce a test which ensures, that this workaround is only applied to Scene Builder 18. If any version number is introduced, there will be a failing unit test with a reminder not towards the related PR.

What do you think? It is still some time till release 19 and a failing test certainly will be a good reminder.

app/src/main/java/com/oracle/javafx/scenebuilder/app/DocumentWindowController.java
```diff
@@ -1573,7 +1573,11 @@ private void performPaste() {
         } else {
             assert isTextInputControlEditing(focusOwner);
             final TextInputControl tic = getTextInputControl(focusOwner);
-            tic.paste();
+            // TODO: Once the issue https://bugs.openjdk.java.net/browse/JDK-8205915
```
jperedadnr, line 1576: Change to https://bugs.openjdk.java.net/browse/JDK-8280057

Oliver-Loeffler: The TODO is now changed to 8280057. How about the switch? Shall I add a failing test for versions > 18? How to go forward here? Once a while, I'll pull OpenJFX to learn but as of now, I'd like to stay with Scene Builder, there's enough to learn at this very moment.

jperedadnr: The issue JDK-8280057 has been closed as duplicated of https://bugs.openjdk.java.net/browse/JDK-8088897 and there might be a fix "soon" for it with https://github.com/openjdk/jfx/pull/694, which is possibly targeting JavaFX 19.

For now I guess there is no other way than apply your patch, and simply keep an eye on the fix on OpenJFX. Once it is integrated and released, we can decide if we need to apply a condition based on a given version (i.e. 18.0.1, 19,  or 19-ea+3).

Adding a unit test might be hard. A system/manual test might work, but we don't have them in Scene Builder.

Oliver-Loeffler: We could create a Scene Builder 19 milestone and assign an issue to it to remove the fix as soon as OpenJFX has received an appropriate fix. @AlmasB What do you think about it? This would not require any further code and we could track it.

AlmasB: Just quickly reading between the lines here.. If I understood correctly, this fix is temporary and will break when openjfx gets a proper fix?

If we are to merge, we do need to keep track of this and possibly keep delaying from milestone to another until the fix is there -- this sounds like a high cost for maintenance.

Could we not have a preference setting that allows enabling / disabling this at runtime? Then, we can future-proof this and fix appropriately without users having to wait for one.

Oliver-Loeffler: Yes it will break, when JavaFX is fixed.

Would you expose this setting in Preferences dialog?

An alternate option would be to check for the JavaFX version used and disable this behavior when using any JavaFX version >= 19/19ea.

Oliver-Loeffler: If you like @AlmasB, you can assign an issue to me so I'll keep an eye on this item.

Oliver-Loeffler: We could add a 2nd preferences page (Platform specific settings) in the existing preferences window.
There we could offer a toggle such as: `Alternate Text Input Control paste behavior = [x]`. How about that?

Oliver-Loeffler: Hi @AlmasB, there is now a preference in the Preferences dialog where the behavior can be turned off. It is enabled by default on MacOS. Also the toggle is disabled for other operating systems. A tool tip provides short description.

Bildschirmfoto 2022-01-21 um 17 47 32

AlmasB (approved):
I think preferences is the way to go, so users can change if necessary.

app/src/main/resources/com/oracle/javafx/scenebuilder/app/preferences/Preferences.fxml
```diff
@@ -118,40 +119,57 @@
...
+    <RowConstraints maxHeight="-Infinity" minHeight="-Infinity" vgrow="NEVER" />
+    <RowConstraints maxHeight="-Infinity" minHeight="-Infinity" vgrow="NEVER" />
+      <RowConstraints />
+      <RowConstraints />
+      <RowConstraints />
+      <RowConstraints />
+      <RowConstraints />
+      <RowConstraints />
+      <RowConstraints />
+      <RowConstraints />
```
abhinayagarwal, line 173: Has this file been changed using Scene Builder?

app/src/main/java/com/oracle/javafx/scenebuilder/app/preferences/PreferencesWindowController.java
```diff
@@ -259,6 +265,15 @@ protected void controllerDidLoadFxml() {
...
         wildcardImports.setSelected(recordGlobal.isWildcardImports());
         wildcardImports.selectedProperty().addListener(new WildcardImportListener());
+        
+        if (EditorPlatform.IS_MAC) {
+            // Alternate paste behavior for Text Input Controls
+            alternatePasteBehavior.setSelected(recordGlobal.isAlternateTextInputControlPaste());
+            alternatePasteBehavior.selectedProperty().addListener(new AlternatePasteListener());
+        } else {
+            alternatePasteBehavior.setDisable(true);
+            alternatePasteBehaviorLabel.setDisable(true);
```
abhinayagarwal, line 275: I prefer not showing it on other platforms. Popup doesn't show up on disabled nodes and this could confuse users on other platforms.

app/src/main/resources/com/oracle/javafx/scenebuilder/app/preferences/Preferences.fxml
Oliver-Loeffler: Yes, this file was edited using Scene Builder. But I'll remove the empty RowConstraint items manually.

app/src/main/java/com/oracle/javafx/scenebuilder/app/preferences/PreferencesWindowController.java
Oliver-Loeffler: Okay, agree for that. Will make this invisible for non-Mac platforms.

app/src/main/resources/com/oracle/javafx/scenebuilder/app/preferences/Preferences.fxml
Oliver-Loeffler: This has been cleaned up.

app/src/main/java/com/oracle/javafx/scenebuilder/app/preferences/PreferencesWindowController.java
Oliver-Loeffler: The option is now only visible on Mac.

app/src/main/java/com/oracle/javafx/scenebuilder/app/preferences/PreferencesRecordGlobal.java
```diff
@@ -126,6 +128,8 @@ public String toString() {
 
     private LocalDate lastSentTrackingInfoDate = null;
 
+    private boolean alternatePasteBehavior = false;
```
jperedadnr, line 131: This should probably be  `true`? At least until there is a fix for it in OpenJFX, and we have a release that we can use.

Oliver-Loeffler: Good point, I'll add a platform specific default. Shall there be any constraints or shall it be true for MacOS in any case? (false for other OS).

jperedadnr: @Oliver-Loeffler You might have missed my question above?

Oliver-Loeffler: Hi @jperedadnr , well, I'll rework this, I understand your point and agree. But currently I am lacking the concentration to do this, my mind is blocked by all the stupidity which is ongoing currently - this makes me really sad and no, don't wanna code, it does not help my mind.

jperedadnr: No worries.
Respect!

Oliver-Loeffler: Hi all, I'll conclude this work until end of this week. Sorry for the delay on that.

jperedadnr: Hi @Oliver-Loeffler , if you are still up for it, it would be great if you could finish this PR. We are about to do a RC with Java/JavaFX 19, and it could probably make it in time for the SB 19 release. Thank you!

Oliver-Loeffler: Hi @jperedadnr, yes I am. Will clean up/refactor accordingly. Hope to get it ready tomorrow.

Oliver-Loeffler: Now the alternate paste behavior is active by default on macOS. Thank you for pinging.

jperedadnr (approved):
works great, thanks for looking into this!

Oliver-Loeffler:
Well, that was fast! Thank you.
Will look more often into Scene Builder again.

abhinayagarwal (approved):

PR #433, opened 2022-01-10
Added initial draft of contribution workflow

### LINKED ISSUE #307: Create a CONTRIBUTING doc
We need to be more clear about how people can contribute. We should be very open to issues, PR's and reviews, but the process needs to be clear.
In general, a PR should refer to an issue, and it should contain a test that fails before the patch, and succeeds after the patch. Also, we should provide guidelines for PRs, so that the review process becomes slightly simpler.

AlmasB:
### Issue

Fixes #307 

### Progress

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,13 @@
+## How to Contribute to Scene Builder
+
+Contribution of any form is welcome! Please see the list below on how you can contribute to the project. Once you've decided what you would like to do, let us know about it first in the [Discussions section](https://github.com/gluonhq/scenebuilder/discussions). This is to make sure that the issue you want hasn't already been implemented or being worked on in newer versions. Any new API or changes to existing API should be discussed to avoid inconsistencies.
+
+* Provide or suggest an implementation of an issue from [GitHub Issues](https://github.com/gluonhq/scenebuilder/issues).
+* Proof read the [public documentation](https://github.com/gluonhq/scenebuilder/wiki) for errors, ambiguities and typos.
+* Create an issue or suggest a feature backed up by a use case.
+* Add missing tests.
+
+## Coding Standards
+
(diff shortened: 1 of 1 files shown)


Oliver-Loeffler:
Hi @AlmasB, this is a very good starter. I am not sure yet how to name it, but we should add a section in limitations as well. There are PRs and Issues where it is asked to add/integrate 3rd party libraries to a certain extent. This can be a tricky thing and it should be clarified if this is indeed something desireable.

AlmasB:
3rd party libraries: I don't think they should be present in `kit` or `app`. In the future, perhaps there can be a separate module, for example `plugins`, that would allow easy extension of SB with 3rd party libraries. However, in the meantime I will not be considering any such extensions, unless approved by other maintainers.

In terms of documentation, I would have thought what you describe falls under `Any new API or changes to existing API should be discussed to avoid inconsistencies.`. So PRs that do not target an issue that is approved for development would be unlikely to get merged. It's a balancing act I suppose: we do not wish to have PRs that are not aligned with the project goal, but at the same time we do not want to discourage people from contributing.

In light of this, perhaps there should be a project aim, or milestone aim, document for the near future -- a strategy of sorts that would underpin major decisions.

Oliver-Loeffler:
Very well said @AlmasB , I really like the idea of having a plugin interface to allow use of external libraries.
Will be the goal of Scene Builder? What will be the vision of this project? Very good question indeed.
As it can be hard to decide what a goal for Scene Builder can be, we could collect desired key properties of Scene Builder.

AlmasB:
I propose this short-term project goal. For the next months, until April 2022, this project will aim:

1. to add more tests.
2. to clean up existing code with a focus on readability, null safety and other improvements to facilitate future maintenance and reduce "entry cost" for new contributors.

During this period, I will refer new features and bug fixes to other project maintainers, as I cannot make informed decisions on these just yet.

johanvos (approved):
Concise and clear, good job.

johanvos:
I agree with the limitations on third party libraries. It's easy to add a 3th party library, but it is hard to maintain that dependency.

Issue #445, opened 2022-01-13
Refactor "app" module


AlmasB:
* Remove extra spaces
* Do not return or accept null in public API

AlmasB:
Also see https://github.com/gluonhq/scenebuilder/pull/447#issuecomment-1018467882

AlmasB:
See https://github.com/gluonhq/scenebuilder/pull/447#discussion_r784180068

AlmasB:
See https://github.com/gluonhq/scenebuilder/pull/447#issuecomment-1019550378

PR #447, opened 2022-01-13
perf: Improve startup time

### LINKED ISSUE #442: Reduce startup times for improved User Experience
Currently the majority of startup time is spent on preparing the first DocumentWindowController (~60% of time) and WelcomeDialogWindow (~40% of time).

The goal is to throw everything we have at getting the WelcomeDialog open as fast we as can for snappy UX, related to #440 

This can be achieved by the following steps:

- [ ] Do not load anything from the file system for WelcomeDialog
- [ ] Pre-bake fxml and CSS into WelcomeDialog code to avoid any parsing
- [ ] Load first DocumentWindowController in the background and only show on demand (that is when the user selects "new" or "open" project)

If the above works as expected, we should be able to get a nice 70-90% boost in startup times.

### LINKED ISSUE #463: Welcome dialog should remain open when open project dialog is cancelled
See discussion in #447 for more details

AlmasB:
Improve startup time by delaying empty document creation and loading recent items in the background.

#442 provides a video demo for the sequence: first we only open the `WelcomeDialog` and defer document window while the user selects their option

### Issue

Fixes #442 
Fixes #463 

### Progress


AlmasB:
It would also be a good idea to test this PR on a different machine (possibly a non-windows OS) to see how improvements fare there...

app/src/main/java/com/oracle/javafx/scenebuilder/app/welcomedialog/WelcomeDialogWindowController.java
```diff
@@ -101,38 +102,51 @@ protected void controllerDidLoadFxml() {
...
+            if (recentItems.size() == 0) {
+                Label noRecentItems = new Label(I18N.getString("welcome.recent.items.no.recent.items"));
+                noRecentItems.getStyleClass().add("no-recent-items-label");
+
+                Platform.runLater(() -> {
+                    recentDocuments.getChildren().add(noRecentItems);
+                });
+            }
+
+            for (int row = 0; row < preferencesRecordGlobal.getRecentItemsSize(); ++row) {
```
cbm64chris, line 128: Could be considered syntactic preference, but this may be cleaner and reduces the size of the loadAndPopulateRecentItemsInBackground method.

```Java
IntStream.range(0, preferencesRecordGlobal.getRecentItemsSize())
    .mapToObj(i -> recentItems.get(i))
    .forEach(this::addRecentDocument);
```

```Java
private void addRecentDocument(String recentItem) {
    File recentItemFile = new File(recentItem);
    String recentItemTitle = recentItemFile.getName();
    Button recentDocument = new Button(recentItemTitle);
    recentDocument.getStyleClass().add("recent-document");
    recentDocument.setMaxWidth(Double.MAX_VALUE);
    recentDocument.setAlignment(Pos.BASELINE_LEFT);
    recentDocument.setOnAction(event -> fireOpenRecentProject(event, recentItem));
    recentDocument.setTooltip(new Tooltip(recentItem));

    Platform.runLater(
         () -> {
            recentDocuments.getChildren().add(recentDocument);
        });
    }
```

Oliver-Loeffler:
Will See to obtain some numbers from MacOS and Ubuntu.

app/src/main/java/com/oracle/javafx/scenebuilder/app/welcomedialog/WelcomeDialogWindowController.java
AlmasB: Looks good, will refactor in a future PR, but not this one, for easier review.

AlmasB:
Yes it's a good idea, but probably for a separate PR. Opened as #450

AlmasB:
- start:
https://github.com/gluonhq/scenebuilder/blob/38e2440657ad332c803e64b1d95006eb37d2f92e/app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java#L120-L123

- end:
https://github.com/gluonhq/scenebuilder/blob/38e2440657ad332c803e64b1d95006eb37d2f92e/app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java#L429-L432

Oliver-Loeffler:
Thanks for point the locations out. So eventually the numbers for mil old 2015 MBP remain. 1.5s with the old version, 1.2s with the updated start process.

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
```diff
@@ -412,31 +412,19 @@ public void handleLaunch(List<String> files) {
...
-                    }
-                });
-            });
-
             // Unless we're on a Mac we're starting SB directly (fresh start)
             // so we're not opening any file and as such we should show the Welcome Dialog
             WelcomeDialogWindowController.getInstance().getStage().show();
 
+            // let JavaFX handle above call ASAP and delay empty document window for improved UX
+            makeEmptyDocumentWindowOnNextTick();
```
abhinayagarwal, line 427: The method name is confusing (to me). 

Can we rename it to `createEmptyDocumentWindowOnNextPulse` ?

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
```diff
@@ -412,31 +412,19 @@ public void handleLaunch(List<String> files) {
...
-            final DocumentWindowController newWindow = makeNewWindow();
-            newWindow.updateWithDefaultContent();
-            newWindow.openWindow();
-
             // Show ScenicView Tool when the JVM is started with option -Dscenic.
             // NetBeans: set it on [VM Options] line in [Run] category of project's Properties.
             if (System.getProperty("scenic") != null) { //NOI18N
-                Platform.runLater(new ScenicViewStarter(newWindow.getScene()));
+                // commented out until it is implemented
+                //Platform.runLater(new ScenicViewStarter(newWindow.getScene()));
```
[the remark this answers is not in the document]
AlmasB: `newWindow` is not accessible there. If you'd prefer I can move it elsewhere, but given it is irrelevant in the meantime, commenting it out seems the most practical option.

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
AlmasB: sure, done.

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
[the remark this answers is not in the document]
abhinayagarwal: Ah, I didn't recognise that `newWindow` is not accessible. This change is then OK here.

jperedadnr:
So far it looks good, but while reviewing I found this issue: 

If you click Open project link (bottom-left) from the Welcome dialog, the FileChooser dialog shows up. If you press cancel, it gets closed, but it also closes the welcome dialog, and since now there is no empty workspace, the app fully exits.

Before this PR, after cancelling the FileChooser, the empty workspace was available.

AlmasB:
@jperedadnr well spotted. I think the underlying issue is with the file chooser logic. It shouldn't close the Welcome dialog if nothing was selected, right?

If so, happy to address that first.

jperedadnr:
Yes, Welcome dialog should remain open.

cbm64chris:
Confirmed. I've always hated the app closing when there is no project window open. I think it should be a specific action such a many macOS app do. I know that is not the default behavior of other OS though.

jperedadnr:
@AlmasB 
Doing some tests, I have some custom controls (quite a few) and in my case this single call
`PreferencesController.getSingleton().getMavenPreferences()` in `SceneBuilderApp:handleLaunch` is taking more than 12 seconds, before showing the welcome screen.
Can you see if we can move most of the code in `SceneBuilderApp:handleLaunch` (mainly maven preferences and user library) to a background thread that works while the welcome screen is being displayed?

Also, we need a menu item to show again the welcome screen.

AlmasB:
@jperedadnr 

I'll have a look at preferences and also other startup phases, perhaps we can unify some threads

Re: menu item, opened #470 to be managed in a separate PR

AlmasB:
@jperedadnr 

I don't have any custom controls, so this is a bit of a blind optimisation, but see if this improves startup for you.

I managed to simplify the background thread ordering as well to avoid back and forth between bg and jfx threads, so should improve overall readability.

jperedadnr:
Thanks, I'll check it out. You can easily add custom controls (ControlsFX, TilesFX, ... via Jar/FXML manager)

jperedadnr:
It is much better now. Now the welcome dialog window shows up immediately, but blank, I see the mouse cursor spinning for a couple of seconds, and after that content is displayed and dialog is fully responsive.

AlmasB:
@jperedadnr 

Platform.runLater() needs to be there since makeNewWindow() needs to be called on the fx thread. Not sure how starting in the background solves the issue, perhaps there is a race condition involving preferences...

AlmasB:
@jperedadnr 

I've moved empty doc creation to background., but left the platform run later as it is a requirement. See if this works, meanwhile I'll think about a cleaner way to access preferences (though not for this PR).

jperedadnr:
I don't think the whole `makeNewWindow` needs the FX thread. Only adding the result to the observable list should be done on it. So maybe you can add the usual check:
```
if (Platform.isFxApplicationThread()) {
            windowList.add(result);
        } else {
            Platform.runLater(() -> windowList.add(result));
        }
```

Anyway, I've run on Mac and Windows without Platform.runLater, works fine for me.

AlmasB:
getStage() which calls new Stage() that can break it too, so wrapped both.

eugener:
I suspect this pattern might be used a lot though out the codebase, does it make sense to extract it into a utility method such as:
```java
   void onFx( Runnable runnable) {
      if (Platform.isFxApplicationThread()) {
            runnable.run();
        } else {
            Platform.runLater(runnable);
        }
   }
```

AlmasB:
Yeah, I'll add a note to #445

jperedadnr:
As a side effect of the current changes of this PR, Welcome dialog shows up really fast and the templates are ready immediately. However, the background tasks are still running, and if you select for instance the Empty template, the workspace shows up without any built-in control. Somehow we need to disable the templates selection until background tasks have finished.

AlmasB:
> Somehow we need to disable the templates selection until background tasks have finished.

Yes, I'm happy with that approach. In fact, in a future PR I wanted to delay the template loading anyway, to make welcome dialog `show()` even faster, so I'll just add that as part of this PR. The goal is to open the window quickly and by the time the user decides what they want, we should be able to finish loading the template contents.

AlmasB:
In the end, went for a less intrusive approach of hiding the templates and showing a progress bar.

jperedadnr:
The progress bar is a good solution so far, but it is in the scrollpane content, so it goes to the top of the dialog, which looks ugly. Probably you need to set the height of the vbox and set its alignment to center (it also needs some right padding).

AlmasB:
Try this one, also updated the colour to match the theme.

Oliver-Loeffler:
The new start process looks good. However, I cannot reproduce any of the slowdowns during Scene Builder start. I've now create a user library with 2000 Jars and barely noticed the progress bar.

The slowest system I have is the old 2015 MBP and there I see the progress bar for approx. 0.7 sec. But there seems to be no track/no indication of progress. There is only the empty bar. May be, to the screen a ProgressIndicator with state INDETERMINATE and a scalex/y of 2 would fit better?

jperedadnr:
Orange is good, but we should keep the gradient? We need to unify colors, though, background is #f68a1f

Oliver-Loeffler:
Thats the way it looks:

https://user-images.githubusercontent.com/22102800/150684560-92886083-957d-4a88-968e-692a5edc6c64.mp4

Oliver-Loeffler:
Here an alternative styling using a Progress Indicator:

https://user-images.githubusercontent.com/22102800/150684786-2249a092-d14d-43de-b9b0-4704dc4cdb50.mp4

jperedadnr:
Definitively, progressIndicator looks better than progressBar.

AlmasB:
Updated to match the latest video, with the label being in black -- a bit more visible.

Oliver-Loeffler:
Cool!

WelcomeWorking

Oliver-Loeffler:
I really like the way you've done the layout. Much simpler than I did it. I knew about `.setAlignment(Pos.CENTER)` but did not consider such a simple thing as `label.setTranslateY(30);`. Very nice and it works really well.

```java
private void setupProgressIndicator() {
        templates.addAll(templateContainer.getChildren());

        var label = new Label();
        label.setText(I18N.getString("welcome.loading.label"));
        label.setTranslateY(30);

        var progress = new ProgressIndicator();
        progress.getStyleClass().add("progress");
        progress.setScaleX(3);
        progress.setScaleY(3);

        templateContainer.setAlignment(Pos.CENTER);
        templateContainer.getChildren().setAll(progress, label);
    }
```

jperedadnr:
Well, in fact we are not waiting while the templates load (that takes a few ms), we wait until preferences, custom controls, and other heavy stuff gets loaded and becomes available.
So a better message could be: `loading components...`, `loading libraries...` or `loading preferences...`. 
Ideally we could have a Service to deal with the background tasks, and send those messages as an update of the background process, if possible.

AlmasB:
Updated text to say "Loading Components...".

Happy to refactor to use Service, but as part of a separate PR.

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
```diff
@@ -370,6 +354,46 @@ public void handleLaunch(List<String> files) {
...
+        if (showWelcomeDialog) {
+            // Unless we're on a Mac we're starting SB directly (fresh start)
+            // so we're not opening any file and as such we should show the Welcome Dialog
+            var welcomeWindow = WelcomeDialogWindowController.getInstance();
+            welcomeWindow.getStage().show();
+
+            startupTasksFinished.addListener((o, old, isFinished) -> {
+                if (isFinished) {
+                    Platform.runLater(() -> {
+                        welcomeWindow.showTemplates();
```
jperedadnr, line 368: method reference

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
```diff
@@ -370,6 +354,46 @@ public void handleLaunch(List<String> files) {
...
+                    });
+                }
+            });
+
+            // let JavaFX handle above call ASAP and delay empty document window for improved UX
+            startInBackground("Set up empty doc", () -> createEmptyDocumentWindow());
```
jperedadnr, line 374: method reference

jperedadnr, line 374: "Set up empty document window"

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
```diff
@@ -370,6 +354,46 @@ public void handleLaunch(List<String> files) {
...
+        } else {
+            // Open files passed as arguments by the platform
+            handleOpenFilesAction(files);
+        }
+    }
+
+    /**
+     * Creates and starts a new daemon thread with [taskName] to executive given [task].
+     */
```
jperedadnr, line 383: Update javadoc with startupTasks?

app/src/main/java/com/oracle/javafx/scenebuilder/app/welcomedialog/WelcomeDialogWindowController.java
```diff
@@ -64,6 +68,7 @@
     @FXML
     private Button emptyApp;
 
+    private List<Node> templates = new ArrayList<>();
```
jperedadnr, line 71: it can be final

app/src/main/java/com/oracle/javafx/scenebuilder/app/welcomedialog/WelcomeDialogWindowController.java
```diff
@@ -77,6 +82,10 @@ private WelcomeDialogWindowController() {
         sceneBuilderApp = SceneBuilderApp.getSingleton();
     }
 
+    public void showTemplates() {
```
jperedadnr, line 85: Add some javadoc

app/src/main/java/com/oracle/javafx/scenebuilder/app/welcomedialog/WelcomeDialogWindowController.java
```diff
@@ -101,36 +110,67 @@ protected void controllerDidLoadFxml() {
...
+        templates.addAll(templateContainer.getChildren());
+
+        var label = new Label();
+        label.setText(I18N.getString("welcome.loading.label"));
+        label.setTranslateY(30);
+
+        var progress = new ProgressIndicator();
+        progress.getStyleClass().add("progress");
+        progress.setScaleX(3);
+        progress.setScaleY(3);
```
jperedadnr, line 170: Move to css

app/src/main/java/com/oracle/javafx/scenebuilder/app/welcomedialog/WelcomeDialogWindowController.java
```diff
@@ -101,36 +110,67 @@ protected void controllerDidLoadFxml() {
...
+            }
+        }, "Recent Items Loading Thread").start();
+    }
+
+    private void setupProgressIndicator() {
+        templates.addAll(templateContainer.getChildren());
+
+        var label = new Label();
+        label.setText(I18N.getString("welcome.loading.label"));
+        label.setTranslateY(30);
```
jperedadnr, line 165: Move to css

app/src/main/resources/com/oracle/javafx/scenebuilder/app/i18n/SceneBuilderApp.properties
```diff
@@ -504,6 +504,7 @@ welcome.recent.items.header = Recent
 welcome.recent.items.no.recent.items = no recent projects
 welcome.new.project.label = New Project
 welcome.open.project.label = Open Project
+welcome.loading.label = Loading Components...
```
jperedadnr, line 507: Add a todo or a follow-up issue to get the translations for this.

app/src/main/resources/com/oracle/javafx/scenebuilder/app/welcomedialog/WelcomeWindow.css
```diff
@@ -76,4 +76,8 @@
     -fx-text-fill: white;
     -fx-font-size: 1em;
     -fx-padding: 0.1666665em 5px 0.1666665em 7px; /* 2 5 2 7 */
+}
+
+.progress {
+    -fx-progress-color: orange;
```
jperedadnr, line 82: We should unify colors. Left pane uses #f68a1f

jperedadnr (commented):
Some more comments.
There is also a small issue: when selecting the Empty template, the welcome dialog closes, but a visible blank stage with the same size shows up for a few ms, until the workspace starts. I see this both on Mac and Windows.

AlmasB:
> There is also a small issue: when selecting the Empty template, the welcome dialog closes, but a visible blank stage with the same size shows up for a few ms, until the workspace starts.

I do not see this, perhaps you can send me the exact libs / custom controls you have, so I can reproduce?

Re: comments, given some of these comments are refactors, which will be superseded by the introduction of Service-based init, would you prefer to enlarge this PR to also include Service-based approach?

jperedadnr:
On windows I don't have custom controls, and the blank window is quite visible: (see at 0:05)

https://user-images.githubusercontent.com/2043230/150773134-f76452bb-7d19-4061-9118-b27e5907b834.mov

About the comments, there is no big refactoring required, it's mainly code formatting

AlmasB:
The blank window is present on master branch too, so it's not something this PR introduces.

https://user-images.githubusercontent.com/3594742/150774663-2202bc38-e9aa-4bff-b9d6-6bfe9f554ad6.mp4

AlmasB:
We should probably open an issue for it and see if we can track it down.

jperedadnr:
yes, makes sense

AlmasB:
All outstanding comments so far fixed

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
```diff
@@ -112,12 +113,14 @@
...
 
     private static SceneBuilderApp singleton;
-    private static final CountDownLatch launchLatch = new CountDownLatch(1);
 
     private final ObservableList<DocumentWindowController> windowList = FXCollections.observableArrayList();
     private UserLibrary userLibrary;
     private ToolTheme toolTheme = ToolTheme.DEFAULT;
 
+    private final ObservableList<Runnable> startupTasks = FXCollections.observableArrayList();
+    private final BooleanBinding startupTasksFinished = Bindings.isEmpty(startupTasks);
```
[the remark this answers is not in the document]
jperedadnr: @AlmasB this is not addressed yet

AlmasB:
Sorry, could you clarify what is needed?

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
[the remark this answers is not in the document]
jperedadnr: Ypu can do:

```
private BooleanBinding startupTasksFinished;
...
public void handleLaunch(List<String> files) {
...
            startupTasksFinished = Bindings.isEmpty(startupTasks);
            startupTasksFinished.addListener((o, old, isFinished)  -> {
...
```
Or directly:
```
public void handleLaunch(List<String> files) {
...
            final BooleanBinding startupTasksFinished = Bindings.isEmpty(startupTasks);
            startupTasksFinished.addListener((o, old, isFinished)  -> {
...
```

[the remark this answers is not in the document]
AlmasB: Happy to do (1), but not sure about the benefits over the current approach?

As for (2), do we have a guarantee that `startupTasksFinished` alongside its listener won't be GC'ed when the binding goes out of scope?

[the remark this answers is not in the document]
jperedadnr: 1) avoids creating the binding object at the beginning. Note that it might be cases where it is not used at all (when `showWelcomeDialog` is false).
2) I have the same thoughts about the risk of the binging being GC'ed (I had bad experiences in the past with Bindings...), that's why I prefer 1).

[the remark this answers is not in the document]
AlmasB: Ok, that clears it up for (1), will do.

AlmasB:
@jperedadnr done

jperedadnr:
Looks good now.

Before we can go ahead and integrate this, can you do the following test: 
- mvn clean install
- mvn package -f app
- copy `app/target/lib/scenebuilder-17.0.0-all.jar` into your the app folder of your existing Scene Builder 17 installation (do a copy of your exiting scenebuilder-17.0.0-all.jar file before...)
- now from your IDE select an FXML file and open it with Scene Builder.

Does it work?

jperedadnr:
See `AppPlatform:AppNotificationHandler`

To fix the above issue, this seems to work for me:
```
// Open files passed as arguments by the platform
            startupTasksFinished = Bindings.isEmpty(startupTasks);
            startupTasksFinished.addListener((o, old, isFinished) -> {
                if (isFinished) {
                    Platform.runLater(() -> handleOpenFilesAction(files));
                }
            });
```

AlmasB:
Yes, my thoughts exactly, however I just realised a possible edge case when startupTasksFinished is finished before the listener is added. So will add an extra check for that.

AlmasB:
So I did a bit of digging, I found these entry points:

1. cmd line -> `handleOpenFilesAction` -> `performOpenFiles`
2. welcome dialog -> select template -> `loadTemplateInWindow`
3. welcome dialog -> open project -> `performOpenFiles`
4. welcome dialog -> recent project ->  `handleOpenFilesAction` -> `performOpenFiles`

All startup tasks must be finished before any of these branches. It seems like the common method is `performOpenFiles`, except for (2).

I'll do some more digging.

AlmasB:
@jperedadnr 

Code is **not** ready for merge, but ready for testing. It accounts for all of the above pathways + I did your test above and they all work.

AlmasB:
@jperedadnr ok, I think it's ready for review

AlmasB:
Closing and reopening to trigger new PR checks.

jperedadnr:
Looks good. 
A couple of comments:
- If you have a number of custom controls and a long list of SB preferences, the welcome dialog shows up really fast, but the list of recent files is empty, and after some time, it is displayed. We need to provide a better indication that the list is being retrieved.
- If you click on a template before the background tasks have finished, it looks like the welcome dialog is closed and a new empty dialog shows up with just a progress indicator and a label. We could have a smoother transition between both views, for instance keeping the old view in the back, and the new view on top as an overlay, with some small level of transparency, or at least with some indication of the selection done (i.e. opening XXX template or file)
- If you click on Open Project before the background tasks have finished, the progress indicator shows up, and the filechooser is displayed on top (and the progress indicator keeps spinning while there is no selection).  And if user cancels, filechooser gets closed, progress indicator removed and templates show up again. However, this shouldn't happen: while the file chooser shows up, and a file is selected, there is plenty of time for the background tasks to finish. If needed, the progress indicator should be shown _after_ the selection is done.
- Tested opening fxml files from IDE or command line, works fine.

AlmasB:
@jperedadnr 

1. There is now a temporary label that indicates this.
2. The masker is now added as an overlay, rather than a replacement.
3. The masker is shown after the user has selected the project to load.

jperedadnr:
Great, looks much better now.
I noticed that the filechooser is non-modal, so templates can be still selected and even open the workspace, while the filechooser dialog is still open. It is an easy fix making it modal.

AlmasB:
The welcome dialog is now the owner of the file chooser window.

app/src/main/java/com/oracle/javafx/scenebuilder/app/welcomedialog/WelcomeDialogWindowController.java
```diff
@@ -32,38 +32,49 @@
...
 
 import java.io.File;
-import java.util.Arrays;
 import java.util.List;
+import java.util.stream.Collectors;
 
 
 public class WelcomeDialogWindowController extends TemplatesBaseWindowController {
 
+    private StackPane newRoot;
```
jperedadnr, line 66: move down, right after `masker`

app/src/main/java/com/oracle/javafx/scenebuilder/app/welcomedialog/WelcomeDialogWindowController.java
```diff
@@ -141,17 +171,97 @@ public static WelcomeDialogWindowController getInstance() {
...
+        var progress = new ProgressIndicator();
+        progress.getStyleClass().add("progress");
+
+        masker = new VBox(progress, label);
+        masker.setAlignment(Pos.CENTER);
+
+        getScene().setRoot(newRoot);
+
+        rootPane.prefWidthProperty().bind(getScene().widthProperty());
+        rootPane.prefHeightProperty().bind(getScene().heightProperty());
```
jperedadnr, line 263: I don't understand why you do a binding here for `rootPane`. Probably better to do it for `newRoot`?  Also, when masker is removed, shouldn't you unbind?

app/src/main/java/com/oracle/javafx/scenebuilder/app/welcomedialog/WelcomeDialogWindowController.java
```diff
@@ -141,17 +171,97 @@ public static WelcomeDialogWindowController getInstance() {
...
+                    // restore root in case welcome dialog is opened again
+                    rootPane.setDisable(false);
+                    newRoot.getChildren().remove(masker);
+                });
+            }
+        });
+    }
+
+    private void initMasker() {
+        newRoot = new StackPane();
```
jperedadnr, line 247: I guess this is okay for now, but Scene Builder tends to build itself from FXML files. It might be easier to have the masker already as an existing invisible node in `WelcomeWindow.fxml`, that can get activated as needed.

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
```diff
@@ -366,10 +351,68 @@ public void start(Stage stage) throws Exception {
...
+                            try {
+                                latch.await();
+                            } catch (InterruptedException e) {
+                                Logger.getLogger(getClass().getName()).log(Level.SEVERE, "An exception was thrown:", e);
+                            }
+                        }
+
+                        Platform.runLater(() -> handleOpenFilesAction(files));
+                    }
+            ).start();
```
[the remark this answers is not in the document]
AlmasB: In this case, I think no. As far as I understand this is the only alive thread and there are no windows open yet, so we have to keep JVM from shutting down.

app/src/main/java/com/oracle/javafx/scenebuilder/app/welcomedialog/WelcomeDialogWindowController.java
AlmasB: `rootPane` is BorderPane (which comes from fxml). Since it is no longer root (`StackPane newRoot` is when the masker is added), it does not grow to fill the parent (the StackPane). The above binding is the easiest way I found to make `rootPane` (and therefore `newRoot`) to fill the parent.

This is not related to masker per se, so no unbinding is needed.

app/src/main/java/com/oracle/javafx/scenebuilder/app/welcomedialog/WelcomeDialogWindowController.java
AlmasB: Given the masker does not need to be initialized for all cases, I think the lazy init is fine here.

AlmasB:
@jperedadnr There are 3 unresolved comments for you to check, the others should be resolved.

AlmasB:
@jperedadnr masker UI is now in fxml

jperedadnr (approved):
It looks really good now, great job

jperedadnr:
I'm in favour of merging and getting some testers to try the snapshot.

AlmasB:
@abhinayagarwal Jose and I have now completed this, please do a quick sanity check in case we missed anything and I'll merge.

abhinayagarwal (approved):
I did a quick sanity check and it seems to be working fine 👍

AlmasB:
@abhinayagarwal thanks, please can you check what's happening with the early release [workflow](https://github.com/gluonhq/scenebuilder/actions/runs/1806606466). It seems the windows package signing is crashing. Once the workflow is green, we can merge this, so that early release artifacts are produced and are testable.

abhinayagarwal:
@AlmasB The [EA workflow](https://github.com/gluonhq/scenebuilder/actions/workflows/early-access.yml) is fixed now!

PR #458, opened 2022-01-16
Configure rolling releases

### LINKED ISSUE #446: Development build releases
It would be beneficial for users to be able to access latest development builds and test them immediately after a merge to `master`. I can see some GitHub actions that generate bundles. Could these be duplicated / extended to upload SNAPSHOT bundles to GitHub or other hosting? @abhinayagarwal

aalmiray:
### Issue

Configures rolling releases on every push to `master` branch.
Configures final release on workflow dispatch (manual trigger)

Rolling releases publish binaries **only** as Git release assets.
Final release additionally publishes binaries to S3.

Fixes #446

### Progress


aalmiray:
Converting to draft as JReleaser `1.0.0-M1` is needed.

pom.xml
```diff
@@ -89,8 +95,193 @@
...
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+
+            <plugin>
+                <groupId>org.jreleaser</groupId>
+                <artifactId>jreleaser-maven-plugin</artifactId>
+                <version>0.10.0</version>
+                <!--version>1.0.0-M1</version-->
```
AlmasB, line 151: I assume this will be updated soon?

pom.xml
```diff
@@ -16,11 +16,17 @@
...
+        <java.version>11</java.version>
+        <maven.compiler.source>${java.version}</maven.compiler.source>
+        <maven.compiler.target>${java.version}</maven.compiler.target>
+        <maven.compiler.release>${java.version}</maven.compiler.release>
         <javafx.version>17</javafx.version>
         <aether.version>1.1.0</aether.version>
         <charm.glisten.version>5.0.0-jdk9</charm.glisten.version>
-        <maven.compiler.release>11</maven.compiler.release>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <project.github.repository>aalmiray/scenebuilder</project.github.repository>
```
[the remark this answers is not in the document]
aalmiray: Yes, it should. I miss this update from my tests

pom.xml
aalmiray: Current plan is to release 1.0.0-M1 on Jan 22/23, so yes, pretty soon.

AlmasB:
Many thanks for this. I've skimmed through to get a rough idea and left some questions, but I don't have the required SB build / release expertise, so I'm tagging @jperedadnr and @abhinayagarwal 

Once this is ready and merged, we should have a much quicker response iteration between the dev team and the users.

abhinayagarwal:
I did an initial overview of the PR. Overall, its great work. Thank you @aalmiray  for taking the time.

Here are my initial thoughts & questions on the PR:

* It would be nice if we could also upload the final release artifacts to homebrew, snap etc. using JReleaser. May be something for a follow-up PR
* Github's limitation of uploading all artifacts in a single ZIP can be frustrating for a user, as they would have to download all the artifacts when their intention would be to just test one of them
* Artifacts are uploaded from individual OS specific workflows. What exactly is JReleaser doing in early-access workflow?
* We might want to replace `TAG` keyword with something else as we no longer do releases via TAG. Does JReleaser pushes a TAG to the repository after a release?

aalmiray:
> It would be nice if we could also upload the final release artifacts to homebrew, snap etc. using JReleaser. May be something for a follow-up PR

Homebrew can be added easily however be aware that this is via a tap repository under your control and not a core formula.

Other packagers such as Snap, Chocolatey, etc are not yet supported. That's something I can have a look at.

> Github's limitation of uploading all artifacts in a single ZIP can be frustrating for a user, as they would have to download all the artifacts when their intention would be to just test one of them

Only if they download artifacts from the GH workflow download area which is not ideal, that's one of the reasons for crafting early access releases, which answers the next question

> Artifacts are uploaded from individual OS specific workflows. What exactly is JReleaser doing in early-access workflow?

JReleaser tags, creates a Git release with changelog, uploads all artifacts as release assets alongside their checksums.

> We might want to replace TAG keyword with something else as we no longer do releases via TAG. Does JReleaser pushes a TAG to the repository after a release?

Right. TAG is akin to the project version while VERSION is akin to the app version used by jpackage. Right now both the project version and the git tag match while the app version differs as it can't contain elements such as -RC or -SNAPSHOT.
Yes, JReleaser can create and push the tag however for the release workflow I configured the Maven release plugin to make that happen as it does 2 things: bump version to release, tag, and push changes.

aalmiray:
Force pushed the following changes:
- fixed repository coords in `pom.xml` as noted by @AlmasB 
- removed comments on features from JReleaser `1.0.0-M1`, We're almost there.
- Added homebrew per @abhinayagarwal's suggestion

Note that a **new** secret named `GITHUB_ACCESS_TOKEN` is required. this is a personal access token that can read/write to repositories on the gluonhq organization. Required to push the brew cask to `gluonhq/homebre-tap` which will be created for you by JReleaser. Refer to https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token

aalmiray:
PR is ready for review.

Package managers such as Homebrew, Macports, Chocolatey, Snap have been disabled as they require additional setup such as service accounts and API keys. We can discuss which package managers should be enabled and their requirements.

For the time being this PR will enable rolling releases on every push to the `master` branch.

.github/workflows/bundles-kit.yml
```diff
@@ -0,0 +1,60 @@
...
+  workflow_call:
+    inputs:
+      version:
+        required: true
+        type: string
+      tag:
+        required: true
+        type: string
+      java-version:
+        default: '17-ea'
```
johanvos, line 13: shouldn't this be 17.0.2?

.github/workflows/bundles-linux.yml
```diff
@@ -0,0 +1,113 @@
...
+            --java-options '"--add-opens=javafx.fxml/javafx.fxml=ALL-UNNAMED"' \
+            --linux-menu-group '"Development;Building;GUIDesigner;Java;"' \
+            --resource-dir app/assets/linux \
+            --install-dir /opt \
+            --type rpm
+          mv ${{ env.INSTALL_DIR }}/*.rpm ${{ env.INSTALL_DIR }}/SceneBuilder-${{ env.TAG }}.rpm
+          ls ${{ env.INSTALL_DIR }}
+        env:
+          MAIN_CLASS: com.oracle.javafx.scenebuilder.app.SceneBuilderApp
+          JAVAFX_HOME: /tmp/javafx-jmods-17/
```
johanvos, line 101: here and below, I see a number of references to 17. We can either use 17.0.2 or 18-ea. 
The next release of Scene Builder will be when JavaFX 18 is released, so maybe we should already depend on 18-ea?

.github/workflows/bundles-kit.yml
aalmiray: Copied the value from the previous workflow. I can move it to `17.0.2` 😄

.github/workflows/bundles-linux.yml
aalmiray: This value could also be made configurable form the caller workflow, just like `JAVA_VERSION` is.

aalmiray: @johan recall it's not just the JAVAFX_HOME directory that needs update but also the URL used to download the mods.

abhinayagarwal: I would configure `javafx-version` as an input in `bundles-kit.yml` just like `java-version`. 

Download link can then it. Linux example:

> https://download2.gluonhq.com/openjfx/{javafx-version}/openjfx-{javafx-version}_linux-x64_bin-jmods.zip

abhinayagarwal:
> Right. TAG is akin to the project version while VERSION is akin to the app version used by jpackage. Right now both the project version and the git tag match while the app version differs as it can't contain elements such as -RC or -SNAPSHOT.

That's correct difference b/w TAG and VERSION. `jpackage` doesn't support -RC or -SNAPSHOT in version.

My initial concern was in case of early-access, we still use TAG, which is coming from previous workflows. However, since there is no tag involved, we may want to rename it to `PROJECT_VERSION` ? And subsequently, `VERSION` to `APP_VERSION` ?

aalmiray:
OK but be aware that changes to the jpackage scripts are also required as they expect `VERSION`.

aalmiray:
I'd recommend squashing commits when merging this PR.

AlmasB:
Yes, squash and merge is the default option for this repo.

johanvos (approved):

abhinayagarwal:
Hi @aalmiray ,

As [suggested](https://github.com/gluonhq/scenebuilder/pull/458#issuecomment-1016271314), I tried adding `GITHUB_ACCESS_TOKEN` to the list of secrets. However, it failed with the following message:

> Failed to add secret. Secret names must not start with GITHUB_.

P.S. We generally use `PAT` as the keyword in other repositories where personal access token is required.

aalmiray:
Oops you are right. I used `GIT_ACCESS_TOKEN` in my repos. Will change to `PAT`.

abhinayagarwal:
@gluon-bot's PAT added as a secret

abhinayagarwal (approved):

aalmiray:
Looks like early-access.yml#L99 must be deleted. I missed it. 

The bundle-kit.yml workflow does not have an "app-version" input.

PR #481, opened 2022-01-21
Fix: Removes white background from register and about modals

gargoyle:
It seemed to me that the "About Scene Builder" and "Register" modal windows didn't look quite right with the white backgrounds when almost all the others seem to have the default "off-white" shade. So this removes the deliberate setting of them to white.

Before:-
image

After:-
image

I also re-saved the logo to use one that was transparent and 2x the size it needed to be, then constrained the image to 78 pixels. Not sure if there is any other special magic trickery that is required for hi-dpi screens?

Finally, Scene Builder updated those `xmlns` versions from 8 to 17 automatically, not something I manually did. I have no clue if that's a problem or not.

### Issue

Issue: #507 

### Progress


abhinayagarwal:
> Finally, Scene Builder updated those xmlns versions from 8 to 17 automatically, not something I manually did. I have no clue if that's a problem or not.

It isn't a problem. FXMLs are automatically updated to the version of the JavaFX, Scene Builder is bundled/running with.

> Issue: None

Please log an issue explaining the incorrect behavior and then link it to the PR.

gargoyle:
Done and edited the first post to include the reference.

PR #487, opened 2022-01-23
Adds clarification about whitespace to the PR template

gargoyle:
When I made my first PR, I interpreted the first item in the list below as needing to remove extra whitespace from any changed files. However, I now think this was meant to mean "don't include a bunch of whitespace only changes".

I think this change helps to clarify that - if I am right in my new interpretation.

### Progress

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -7,6 +7,6 @@ Fixes #
 
 ### Progress
 
-- [ ] Change must not contain extraneous whitespace
+- [ ] Change must not contain extraneous whitespace changes
 - [ ] License header year is updated, if required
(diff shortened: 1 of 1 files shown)


abhinayagarwal:
> When I made my first PR, I interpreted the first item in the list below as needing to remove extra whitespace from any changed files.

It already says that "Change must not..". Isn't adding 'changes' at the end redundant?

AlmasB:
Thanks for your perspective on this, but I'm with @abhinayagarwal on this one. The term "change" typically does not refer to a file, but refers to the set of differences (particularly the `++`) between the two versions: the old and the new. Therefore, I believe "Change must not contain extraneous whitespace" parses fine, meaning the `++` (added) code must not contain extra empty lines.

If the above makes sense, I'm voting to close this PR. Alternatively, happy to discuss further how to clarify the template wording.

gargoyle:
Fair enough. :-)

PR #516, opened 2022-02-07
refactor: Removed null field and logic from Template enum

AlmasB:
### Issue

Fixes #514 

### Progress

- [x] Change must not contain extraneous whitespace
- [x] License header year is updated, if required
- [x] The PR name must follow the [pre-defined format](https://github.com/gluonhq/scenebuilder/blob/master/CONTRIBUTING.md)
- [x] Verify the contributor has signed [Gluon Individual Contributor License Agreement (CLA)](https://docs.google.com/forms/d/16aoFTmzs8lZTfiyrEm8YgMqMYaGQl0J8wA0VJE2LCCY)

PR #518, opened 2022-02-18
feat: improve welcome dialog UX

### LINKED ISSUE #475: Improve Welcome UX
Currently, "new" and "open" are implemented as radio buttons, with "new" being redundant since the new projects list is always visible. Hence,

- [ ] Remove "new" button for clearer UX
- [ ] Turn "open" into a normal button
- [ ] The window should not be resizable beyond its min size of contents
- [ ] There is a subtle focus issue (see beginning of video) due to ScrollPane (both left and center panes) being focusable. This needs to be fixed.

https://user-images.githubusercontent.com/3594742/150533038-4599ef64-8eb9-40e9-980a-c1b28615b193.mp4

AlmasB:
### Issue

Fixes #475 

Specifically, these are implemented:
- [x] Remove redundant "new" button for clearer UX
- [x] Turn "open" into a normal button
- [x] The window should not be resizable beyond its min size of contents
- [x] There is a subtle focus issue (see beginning of video in #475) due to ScrollPane (both left and center panes) being focusable. 
- [x] `Open Project' is centered
- [x] Templates don't have a title. Probably a top bar with brown background and the old title "New Project" would help (scrollpane below it, so it is always visible)
- [x] Templates labels ("Empty", "Basic Application",...) are maybe too small, and black font color could be probably close to the brown one.
- [x] Templates have a description below their title.
- [x] The "buttons" on the dialog (templates, recent, open) don't show a hand cursor, so it is not clear that those could be clicked.
- [x] Empty is now part of Desktop, since Mobile already has one.

### Progress


AlmasB:
New view for reference
Screenshot 2022-02-18 10 26 00

AlmasB:
Good suggestions, now updated.

Screenshot 2022-02-19 20 43 31

app/src/main/java/com/oracle/javafx/scenebuilder/app/welcomedialog/WelcomeDialogWindowController.java
```diff
@@ -170,6 +170,14 @@ private void loadAndPopulateRecentItemsInBackground() {
         t.start();
     }
 
+    private void setupTemplateTooltips() {
+        Tooltip.install(emptyApp, new Tooltip(I18N.getString("template.description.new.empty.app")));
```
eugener, line 174: For maintainability, it might be better to use a loop through `apps` and related `templates`

AlmasB: Actually, I've been thinking to add the tooltip text to #516 

It seems each _app_ button has `userData = template`, so then the loop becomes trivial. Happy to refactor here, but #516 should be merged first to avoid merge issues.

eugener: Also, in my opinion, there is enough space for template descriptions to be presented permanently, just below the title. The title can use a bold(er) font and wrapped text below with regular one, for example

AlmasB:
I wonder if the Empty template should be under Desktop, since Phone already has one...

AlmasB:
Moved Empty to Desktop. Set template descriptions to always be visible.

Screenshot 2022-02-20 17 32 55

eugener (dismissed):
Looks a lot better IMO - informative at a glance. 
More UX iterations can be done later if needed.

jperedadnr:
Looks good. A few comments:

- Label descriptions are out of hovered template:

image

- `New Project from Template` font size is smaller than `Open Project`, which is also smaller than `Recent`.
- `Recent` could probable be renamed to `Open Recent Project`? It also needs some alignment with the other two titles.

AlmasB:
I started making the above changes, then discovered that we also have a menu item "New from Template", which opens a separate template window with a different set of fxml/css. To avoid keeping two separate sets of files to be updated, this change now also gets rid of code/data duplication by extracting `TemplatesWindow.fxml` from `WelcomeWindow.fxml` and includes it via `fx:include`.

Along similar lines, to avoid making changes separately to each template in the UI, the templates are now generated dynamically from the `Template` enum. Adding / removing templates should no longer require changes to the UI or logic.

The separate template window now behaves exactly the same (it _is_ the same thing) as the welcome window's template section:

Screenshot 2022-02-21 17 16 26

AlmasB:
@jperedadnr all ready here, if you'd like to review so we can get this in for SB18

jperedadnr:
Minor issues pending:

- Padding required:
 
image

Ideally add padding so Top-pane StackPane has same height as Open Project button (now 23 vs 36). 

Also background color is different (0xd67718 vs 0xa2642a)

-Padding or wrap text

image

- Recent Projects could be renamed to `Open Recent Project`?

- Indent recent project names? 

image

- After showing tooltip over a recent project, mouse cursor goes back to the default pointer instead of keeping the hand cursor.

And about the "New from templates" dialog:
- Dialog title is "Templates", probably change to Gluon Scene Builder (same as welcome dialog).
- Maybe add same top pane stack pane and add label "New Project from Template"?
- Same issue with padding around each template.

jperedadnr:
Okay with the different colors. Anyway some padding is needed. About not being a button, that's the risk of styling buttons as labels, the only thing that changes is the cursor when you hover them.

Mouse cursor: Seems a weird issue with `-fx-cursor: hand` in the templates. Could be a bug like https://bugs.openjdk.java.net/browse/JDK-8179379, so no worries for now.

AlmasB:
Done, including some (not necessarily matching) padding. 

See what you think

jperedadnr:
Looks much better 👍 

Other issue I just noticed: the Open from Template dialog doesn't get closed when selecting a template, is that on purpose?

AlmasB:
I don't think this PR affects the logic behind the open from template dialog, but I think it should close, so I will add that change in and also add min size constraints for the Stage, as with Welcome Dialog.

AlmasB:
@jperedadnr the two comments above addressed: min size and closing stage after template selection.

jperedadnr:
Another minor issue: the VBox is not resized and it shows some empty space to the right.

image

AlmasB:
Odd, I don't have it on Windows 10. Resizing does not reproduce...

Screenshot 2022-03-02 20 35 53

jperedadnr:
You set in `com.oracle.javafx.scenebuilder.kit.template.TemplatesWindowController`:
```
getStage().setMinWidth(800);
```
but then `TemplatesWindow.fxml` has:

```
prefWidth="791.0"
```

Removing the former set fixes it.

AlmasB:
Should we set the stage min width to 750 then?

kit/src/main/java/com/oracle/javafx/scenebuilder/kit/template/TemplatesBaseWindowController.java
```diff
@@ -84,28 +73,52 @@ protected void controllerDidCreateStage() {
...
+    private void initTemplates() {
+        FlowPane desktopPane;
+        FlowPane mobilePane;
+
+        try {
+            ScrollPane templatesRoot = (ScrollPane) getRoot().lookup("#templatesRoot");
+            desktopPane = (FlowPane) templatesRoot.getContent().lookup("#desktopPane");
+            mobilePane = (FlowPane) templatesRoot.getContent().lookup("#mobilePane");
+        } catch (Exception e) {
+            System.out.println("Failed to lookup() [desktopPane] and [mobilePane]: ");
```
jperedadnr, line 92: We should use the logger for all printouts, and at least every new PR should use it.
We shouldn't print stack traces to the console either. This is only visible if user runs from command line. 
We should log the error and the exception to the logs, which are visible to anyone.

kit/src/main/java/com/oracle/javafx/scenebuilder/kit/template/TemplatesBaseWindowController.java
```diff
@@ -84,28 +73,52 @@ protected void controllerDidCreateStage() {
...
+                Label label = (Label) btnRoot.lookup("#labelDescription");
+                label.setText(template.getDescription());
+
+                if (template.isDesktop()) {
+                    desktopPane.getChildren().add(btnRoot);
+                } else {
+                    mobilePane.getChildren().add(btnRoot);
+                }
+            } catch (Exception e) {
+                System.out.println("Failed to load template: " + template);
```
jperedadnr, line 118: Same comment

jperedadnr (commented):
Looks good. Minor comments about logging.

kit/src/main/java/com/oracle/javafx/scenebuilder/kit/template/TemplatesBaseWindowController.java
AlmasB: Good idea, I'll fix these here and open an issue for others in the codebase.

AlmasB:
@jperedadnr fixed sysout calls and created #526

jperedadnr (approved):
looks good!

PR #519, opened 2022-02-20
fix: Adds status icon to the Check for Updates menu item. Fixes #489

### LINKED ISSUE #489: Update dialog should be extracted into a non-intrusive notification
### Expected Behavior

Update notification is shown elsewhere in the SB workflow

### Current Behavior

Update dialog is shown if necessary at the end of welcome dialog workflow

gargoyle:
### Issue

Fixes #489

Adds an icon to the "Help -> Check for Update..." menu and displays in one of the following states:-

All OK and up-to-date:-
image

Update is available:-
image

Exception thrown in the existing version checking code:-
image

### Progress


AlmasB:
Looks great!

It seems there are 3 cases: OK, update available, exception. I think we can simplify this for the user:

**case exception:** (the user doesn't need to know there is an error, so just fall through to case OK)
**case OK:** we should show OK icon and say "All up-to-date!"

**case update available:** we should show download icon and say "Update available..."

gargoyle:
I'm not sure about doing it all from the FXML, I might need some help with that.

It looks like the existing logic handles this with the `SceneBuilderApp::canPerformControlAction()` method. By getting the switch to return false for `CHECK_UPDATES` , the menu item becomes disabled:-
image

If I were to change this one to being two menu items that are shown/hidden, it would make it the odd one out.

I can easily add the same `if` check for the version to this switch statement, but I'll have a stab at refactoring a little bit (rule of thirds and all that).

AlmasB:
You could go with "All OK" in FXML with the appropriate icon. Then, in code, check for updates and overwrite if an update is available.

Let me know if you need help with the FXML part of it.

gargoyle:
Oh, thanks @jperedadnr. I'll have a play round.

jperedadnr:
Seems that [`GlassSystemMenu::getPixels`](https://github.com/openjdk/jfx/blob/master/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassSystemMenu.java#L292) doesn't check other than the url of the image to get the pixels down to native. 
So you'll need to provide a smaller size for the icons?

AlmasB:
My original proposal was to use something like `new red Circle()`, which is visible when there is an update, otherwise not. Both for the update item and the "Help" menu title. I wonder if that would simplify this issue.

jperedadnr:
That won't work for Mac, see https://github.com/openjdk/jfx/blob/master/modules/javafx.graphics/src/main/java/com/sun/javafx/tk/quantum/GlassSystemMenu.java#L293: only ImageView...

AlmasB:
In which case, for this PR, we could ditch the visual indicator and do:

- If update available, the menu item text is "Update available..."
- Else, the menu item text is "Up to date" and greyed out (disabled).

Later, once we have a notification service of some sort like a non-modal pop-up in the top-right corner or something, we can make use of that to notify the update is available.

gargoyle:
Thanks for all the feedback. 

@AlmasB I agree with the simpler UX for the user, so I removed the exception state, and the default if anything goes wrong is to show the OK message, as you suggested.

However, when simulating a delay in the thread that goes off and grabs the version number from the web, I felt some kind of user feedback that something was happening in the background would be better. So I added an initial state of "Checking for update...". This is what it looks likes with a forced 10s delay (sorry about the washed out gif colours):-

slow_net_sim

Since there is now no image for the default state, I've not added anything else to the FXML. If you still want to experiment with that, then I'll have to take you up on your offer for some FXML assistance because I couldn't figure that bit out.

@jperedadnr I found an [article from Dirk Lemmermann](https://dlsc.com/2017/08/29/javafx-tip-27-hires-retina-icons/) and updated the images and code accordingly. However the same oversized icon was rendered, so this is deffo a bug in the Mac code as it correctly identified a high-res system and loaded `check_icon@2x.png` but then still rendered it at double the size. So I have just removed the high-res versions for now. 

This is how it looks on Mac:-

help_menu_mac

And on Linux:-

help_menu_linux

And on Linux when there's an update available:-

help_menu_linux_update

gargoyle:
FYI: I think I fixed the icon rendering issue in openjfx:- https://github.com/openjdk/jfx/pull/743. So hopefully we can put some high res icons back in play at some point in the future (if/when that PR gets merged).

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
```diff
@@ -235,6 +234,10 @@ public boolean canPerformControlAction(ApplicationControlAction a, DocumentWindo
                 result = true;
                 break;
 
+            case CHECK_UPDATES:
```
AlmasB, line 237: Do we need to make this change at all? If the update is not available, the item is disabled, so not clickable. I think the existing behaviour is fine here.

app/src/main/java/com/oracle/javafx/scenebuilder/app/menubar/MenuBarController.java
```diff
@@ -1171,6 +1176,24 @@ public String getTitle() {
         insertMenu.setOnMenuValidation(onCustomPartOfInsertMenuValidationHandler);
         
         windowMenu.setOnMenuValidation(onWindowMenuValidationHandler);
+
+        // Modify the Check Updates menu item if there is an update available.
+        // Icons by Font Awesome (https://fontawesome.com/license/free) under CC BY 4.0 License
+        AppSettings.getLatestVersion(latestVersion -> {
+            var updateAvailable = AppSettings.isUpdateAvailable();
+            Image icon = new Image(MenuBarController.class
+                    .getResource(updateAvailable ? "download_icon.png" : "check_icon.png")
```
AlmasB, line 1185: I think we can remove `warn_icon` from this PR.

app/src/main/java/com/oracle/javafx/scenebuilder/app/menubar/MenuBarController.java
```diff
@@ -1171,6 +1176,24 @@ public String getTitle() {
...
+                    .toExternalForm());
+            ImageView iconView = new ImageView(icon);
+
+            Platform.runLater(() -> {
+                checkUpdatesMenuItem.setGraphic(iconView);
+                checkUpdatesMenuItem.disableProperty().setValue(!updateAvailable);
```
AlmasB, line 1191: In FXML, I'd disable the item by default and use `menu.title.check.updates.ok` text with `"check_icon.png"` graphic. If you open the .fxml in Scene Builder you can edit the update menu item.

Then the code just checks if the update is available and, if it is, overwrites it with "download_icon" and "menu.title.check.updates.available" text.

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
gargoyle: Yes. The checks are performed every time you open the menu. Even with disabling the menu item in the FXML file, if we leave this hard coded to a `true` response, it will re-enable the item when you open the menu.

app/src/main/java/com/oracle/javafx/scenebuilder/app/menubar/MenuBarController.java
gargoyle: Done. Good catch. :-)

app/src/main/java/com/oracle/javafx/scenebuilder/app/menubar/MenuBarController.java
gargoyle: Thanks for that snippet, my google-fu was failing me. Had to prefix the url with '@' though. 

Pushed changes for all these suggestions now. 👍

app/src/main/java/com/oracle/javafx/scenebuilder/app/menubar/MenuBarController.java
```diff
@@ -56,6 +57,7 @@
 
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
```
AlmasB, line 60: I think we can remove this import.

AlmasB (approved):
Looks good, thanks.

@jperedadnr did you want to have another look at this? If not, we are good to go.

app/src/main/resources/com/oracle/javafx/scenebuilder/app/i18n/SceneBuilderApp.properties
```diff
@@ -191,7 +191,9 @@ menu.title.no.window = No Window
 # Help menu items
 menu.title.scene.builder.help = Scene Builder Help
 menu.title.show.welcome = Show Welcome Page
-menu.title.check.updates = Check for Update...
+menu.title.check.updates.available = Update available
+menu.title.check.updates.ok = All up-to-date!
```
jperedadnr, line 195: This text out of context can mean many things. We need a better wording like "Updates: Scene Builder is up-to-date", which is longer, but I see that the menu is already wider than expected:

image

app/src/main/java/com/oracle/javafx/scenebuilder/app/menubar/MenuBarController.java
```diff
@@ -1171,6 +1175,23 @@ public String getTitle() {
...
+        AppSettings.getLatestVersion(latestVersion -> {
+            if (AppSettings.isUpdateAvailable()) {
+                Image icon = new Image(MenuBarController.class
+                        .getResource("download_icon.png")
+                        .toExternalForm());
+                ImageView iconView = new ImageView(icon);
+
+                Platform.runLater(() -> {
+                    checkUpdatesMenuItem.setGraphic(iconView);
+                    checkUpdatesMenuItem.disableProperty().setValue(false);
```
jperedadnr, line 1190: Does it make sense to disable this menu option? 

In case the user really wanted to be sure, clicking on it should at least bring a dialog saying that the current version XXX is already up-to-date, which is what we already do now?

image

app/src/main/java/com/oracle/javafx/scenebuilder/app/util/AppSettings.java
```diff
@@ -136,6 +140,13 @@ public static void getLatestVersion(Consumer<String> consumer) {
                     ex.printStackTrace();
                 }
                 latestVersion = onlineVersionNumber;
+
+                try {
+                    isUpdateAvailable = isCurrentVersionLowerThan(latestVersion);
+                } catch (NumberFormatException e) {
+                    e.printStackTrace();
```
jperedadnr, line 147: Move this to Logger.logSevere ? We shouldn't print stacktraces to console as they get lost unless you run from command line.

jperedadnr:
@AlmasB I've added some comments.
@gargoyle Looks good, it's also great that you created the OpenJFX PR.

app/src/main/java/com/oracle/javafx/scenebuilder/app/util/AppSettings.java
gargoyle: I was just following what the rest of the code in the block was doing. `e.printStackTrace()` seems to be used quite a lot.

jperedadnr: I understand, but we have logs for a reason, and that's how new PRs should be tackling exceptions.

gargoyle: I did a search of how Logging is being done elsewhere. Seems we have two patterns:-

```
Logger.getLogger(AppSettings.class.getName()).log(Level.SEVERE, "An exception was thrown:", ex);
```

or

```
Logger.getLogger(PreferencesRecordDocument.class.getName()).log(Level.SEVERE, null, ex);
```

With the second one seeming to be a bit more common. Any recommentation on which to use? 

I'm assuming from the output that it will be clear its an exception, making the "An exception was thrown:" text redundant?

AlmasB: The `String` message should normally include some useful information. So in the above case it might say: "Cannot parse version string" or something similar.

app/src/main/resources/com/oracle/javafx/scenebuilder/app/i18n/SceneBuilderApp.properties
gargoyle: How about "No updates available"?

Also, menu doesn't have that extra space on Linux :shrug: 
image

Oliver-Loeffler:
@gargoyle Would you mind to have a look at the conflicts? May be we can get this one eventually merged. It technically looks fine to me.

gargoyle:
I had a quick look this evening and got myself into a rebase nightmare. :roll_eyes: 

I've hardly looked at SceneBuilder (or even done much JavaFX) in the last two years, so what will probably be best is that I'll see if I can find some time in the next week or so to get back into it a bit, and just re-implement this on a fresh branch cut from the latest main/master.

My fix for MacOS icon rendering did make it into JavaFX 20, so if SceneBuilder is targeting that or later as a min requirement, I can redo the high res versions of the icons too?

Will need to pair up with someone though, as I don't even have my old Mac to check even basic stuff anymore.

Oliver-Loeffler:
Hi Paul, I still have my old MacBook so I can help out there. Very cool that you'checked in Here. Thanks!

Am 27. September 2024 20:14:38 UTC schrieb Paul ***@***.***>:
>I had a quick look this evening and got myself into a rebase nightmare. :roll_eyes: 
>
>I've hardly looked at SceneBuilder (or even done much JavaFX) in the last two years, so what will probably be best is that I'll see if I can find some time in the next week or so to get back into it a bit, and just re-implement this on a fresh branch cut from the latest main/master.
>
>My fix for MacOS icon rendering did make it into JavaFX 20, so if SceneBuilder is targeting that or later as a min requirement, I can redo the high res versions of the icons too?
>
>Will need to pair up with someone though, as I don't even have my old Mac to check even basic stuff anymore.
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/gluonhq/scenebuilder/pull/519#issuecomment-2379985981
>You are receiving this because you commented.
>
>Message ID: ***@***.***>
-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

PR #528, opened 2022-03-14
refactor: Replaced print stack trace calls with Logger

### LINKED ISSUE #526: All calls to System.out/err and printStackTrace() should be replaced with appropriate Logger

AlmasB:
### Issue

Fixes #526 

### Progress

### DIFF
--- a/app/src/main/java/com/oracle/javafx/scenebuilder/app/util/AppSettings.java
+++ b/app/src/main/java/com/oracle/javafx/scenebuilder/app/util/AppSettings.java
@@ -50,6 +50,8 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.Properties;
 import java.util.function.Consumer;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 public class AppSettings {
     public static final String APP_ICON_16 = SceneBuilderApp.class.getResource("SceneBuilderLogo_16.png").toString();
@@ -81,8 +83,8 @@ public class AppSettings {
                 sbProps.load(in);
                 sceneBuilderVersion = sbProps.getProperty("build.version", "UNSET");
             }
-        } catch (IOException ex) {
-            ex.printStackTrace();
+        } catch (IOException e) {
+            Logger.getLogger(AppSettings.class.getName()).log(Level.WARNING, "Cannot init SB version:", e);
         }
     }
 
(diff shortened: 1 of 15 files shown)


kit/src/main/java/com/oracle/javafx/scenebuilder/kit/editor/panel/content/mode/EditModeController.java
```diff
@@ -825,7 +829,7 @@ private void keyPressedOnGlassLayer(KeyEvent e) {
                 if (RelocateSelectionJob.isSelectionMovable(contentPanelController.getEditorController())) {
                     activateGesture(new MoveWithKeyGesture(contentPanelController), e);
                 } else {
-                    System.out.println("Selection is not movable");
+                    Logger.getLogger(getClass().getName()).log(Level.INFO, "Selection is not movable");
```
abhinayagarwal, line 832: Same here

app/src/main/java/com/oracle/javafx/scenebuilder/app/util/AppSettings.java
```diff
@@ -160,10 +162,10 @@ private static void updateLatestVersionInfo() {
                 latestVersionText = announcementObject.getString("text");
                 latestVersionAnnouncementURL = announcementObject.getString("url");
             } catch (IOException e) {
-                e.printStackTrace();
+                Logger.getLogger(AppSettings.class.getName()).log(Level.WARNING, "Failed to read latest version json: ", e);
             }
         } catch (MalformedURLException e) {
-            e.printStackTrace();
+            Logger.getLogger(AppSettings.class.getName()).log(Level.WARNING, "Failed to construct latest version info URL: ", e);
```
abhinayagarwal, line 168: Isn't it better to create a single `static final` instance of the Logger for the class?

Oliver-Loeffler: Would only do that if there are multiple calls at various places inside a class.
Would not create that if it is only an exception handling.

Opposite would be for example when logging state of object or node tree modifications, there I'd hold a (`private static final`) logger instance for this particular class.

Oliver-Loeffler: For this class it would be possibly okay to hold a Logger instance. But, I also think this work is important and it's not necessary to be perfect from day one. Actually I'd appreciate if we could get this into Scene Builder 19. It will possibly turn out helpful as we would eventually get some useful logfiles in future.

Also, we could discuss how logging should be handled in future. I offer to run the necessary refactorings to implement the rules which we possibly might agree upon. But as of now, this PR will already be quite useful if merged.

jperedadnr (approved):
looks good to me as is, but of course it can be improved in the future if needed

PR #576, opened 2022-10-03
fix: enable Scene Builder to load FXML with unresolved imports (#120, #281, #733)

### LINKED ISSUE #120: Optimizing FXML imports deletes classes needed in fx:define
Originally reported by: **A. Walter (Bitbucket: [awalterwork](https://bitbucket.org/awalterwork), GitHub: Unknown)**

In **SceneBuilder 8.3.0**, the FXML imports are "optimized" whenever the file is saved: Only imports for the exact classes used in the node tree remain.

However, when you used fx:define to define some objects, SB does not keep the imports for their classes when it saves the file - it just deletes them! (It does not make a difference if you were using wildcards or not in your imports.) 

This means that after editing the file with SB, you cannot load the FXML file in your application anymore - SB effectively breaks it.

# Steps to reproduce: #
* Open the example FXML (see attached files) in the example application (see attached files): it can be displayed.
* Open the example FXML in SceneBuilder.
* Change something (for example, the text of the button in the red area) and save.
* Look at the example FXML source. There are not imports for the classes referenced only in fx:define anymore (ArrayList, SplitMenuButton, MenuItem).
* Try to open the example FXML in the example application: The application cannot start because of LoadExceptions like "javafx.fxml.LoadException: ArrayList is not a valid type.".

# Expected Behaviour: #
- SceneBuilder should save the file in a usable state, that is, with imports for all classes referenced in the FXML file.

- Bitbucket: https://bitbucket.org/gluon-oss/scenebuilder/issue/120

### LINKED ISSUE #281: ClassDefNotFoundError when importing custom component
As [suggested on StackOverflow](https://stackoverflow.com/questions/63325518/gluon-scene-builder-custom-controls-from-source-code?noredirect=1#comment111976855_63325518) I build SceneBuilder from master and tried out the feature apparently merged with #228.

First, everything looked fine but after all, it did not load my custom component. I then debugged it in IntelliJ and found out that I get a `ClassDefNotFoundError` for all .class files in the folder I provided, even though these classes clearly exist.

```
Java
11.0.6+10-LTS, Azul Systems, Inc.

Operating System
Windows 10, amd64, 10.0
```

Does anyone have the same issue?

Plus: I am not sure if it is a good thing that the application fails silently without telling the user why loading a custom component failed.

### LINKED ISSUE #733: build: Tests which ran in the non-modular setup are now failing with the new modular setup
Certain tests, e.g. as provided in PR #576 are now failing with the current setup using Maven-Surefire 3.0.0-M5.

### Expected Behavior
* tests shall pass (as those test passed w/o `module-info.java`)

### Current Behavior
* tests in kit break with an error message about wrong module configuration
```
[ERROR] Errors:
[ERROR]   FXOMImportsRemoverTest.that_fxml_is_not_modified_when_nothing_is_to_be_removed » InaccessibleObject
[ERROR]   FXOMImportsRemoverTest.that_the_given_imports_are_removed » InaccessibleObject
[ERROR]   FXOMRefresherTest » InaccessibleObject Unable to make public static void com.o...
[ERROR]   FXOMSaverTest » InaccessibleObject Unable to make public static void com.oracl...
```

### Steps to Reproduce
* clone Scene Builder from [issue-120](https://github.com/Oliver-Loeffler/scenebuilder/tree/issue-120) and checkout commit 55f95ab49f292761a7dbe5ea7ca47b92ef3d7ea3.
* cd into kit and run `mvnw verify`

### Your Environment
* Windows 22H2
* Java 23
* OpenJFX 23
* Maven 3.9.9

### Screenshots

grafik

grafik

Oliver-Loeffler:
This fix allows Scene Builder to load FXML files with unresolved imports. 
There is also a new option (keep unresolved imports) which also allows to keep the imports in saved FXML files. This can be toggled in Scene Builder preferences dialog. 

If disabled, Scene Builder will effectively omit those unresolved imports and remove them from the updated FXML to be written.
If enabled, Scene Builder will keep track of unresolved imports and write them (unmodified) into the updated FXML. The import remains as it was declared in the original FXML file. Optimization is not applied to unresolved imports.

Of course, one cannot make use of unresolved imports and edit these controls. But, one can at least open an FXML file, edit and rework it in parts. Also, Scene Builder will inform the user that there is something wrong with the imports - it will not silently ignore the issue. This is quite helpful with foreign projects where users only want to look into the files and get a feeling of how the UI would look like. 

This PR also increments Maven-Surefire version from 3.0.0-M5 to 3.5.0 in order to mitigate module system related errors during test.

### Issue

Fixes #120
Fixes #281
Fixes #733 (added on 2024-09-29)

### Progress


Oliver-Loeffler:
This one no longer works as it used to. I'll review and rework accordingly. Loading FXML files with unresolvable imports works. Saving does not work anymore. I'll also add example FXMLs and tests for the FXOM.

Oliver-Loeffler:
@abhinayagarwal Could you please review this one? 
It's not finished yet, but it addresses an actual issue.

The idea is:
* Allow the user to accept (or not) the presence of unresolvable imports via SceneBuilder preferences.
* If it is permitted to load FXMLs with unknown types:
    * The resulting FXOMDocument will keep track of the unknown types.
    * FXOMSaver and FXOMRefresher will ignore those imports (e.g. will also remove them conditionally from any intermediate created FXML text).

It is technically working and as of now I'm moving towards the import replacement logic - this is not robust yet.

Interesting would be your opinion @abhinayagarwal in terms of if this approach is useful.

Thanks!

Oliver-Loeffler:
The way how it is setup now, we would be able to offer 2 options to handle unresolvable FXML imports:
a) Ignore the imports during load but keep them when saving the FXML (preserve)
b) Ignore the imports during load but dont save them at all (remove unresolvable imports)

Actually I can imagine cases for both options. The question is, where to offer those options? All in preferences menu or for case b) as a special save command?

abhinayagarwal:
can there be a condition when the un-resolvable imports of UI classes causes the FXML to not load and show up in SB?

Oliver-Loeffler:
> can there be a condition when the un-resolvable imports of UI classes causes the FXML to not load and show up in SB?

Well, afaik, Scene Builder indicates unknown components with a yellow sign and an exclamation mark. 
I have not yet encountered a FXML file where this would happen.

As of now, (SB21, SB22) Scene Builder cannot load FXMLs with un-resolvable imports. There will be an error message from the FXMLLoader.

abhinayagarwal:
I tested this PR, here are my observations:

1. When SB is unable to find import for a Parent, it will load the FXML with with a yellow sign
2. When SB is unable to find import for a Control, it will fail to load the FXML and show "File Open Error" dialog

I wonder what value 1 brings to the table because any components/children in the parent will be lost if edited in SB 🤔

Oliver-Loeffler:
Please checkout the preferences dialog. There is a new setting.

grafik

Also SceneBuilder will now also notify the user that there are unresolvable imports and that an FXML file with such cannot be properly edited. However, one can at least open it and work with it. Especially since there are cases where the imports are sometimes only needed to get some userData.

grafik

However, default behavior is to not allow loading of FXMLs with unknown or broken imports.

Interestingly, the FXOMDocument knows that there are unknown imports. One could even notify the user upon save action that the resulting FXML might break.

This is IMHO bettern than either not opening an FXML or silently breaking it.

Oliver-Loeffler:
Hi @abhinayagarwal, do you have any chance to review this one again?

kit/src/test/java/com/oracle/javafx/scenebuilder/kit/JfxTestHelper.java
```diff
@@ -0,0 +1,58 @@
...
+    private JfxTestHelper() {
+        /* test utility - not intended for instantiation */
+    }
+    
+    public static <T> T waitFor(Callable<T> callable) throws Exception {
+        FutureTask<T> task = new FutureTask<T>(callable);
+        if (Platform.isFxApplicationThread()) {
+            return callable.call();
+        } else {
+            Platform.runLater(()->task.run());
```
jperedadnr, line 53: `Platform.runLater(task);`

kit/src/test/java/com/oracle/javafx/scenebuilder/kit/fxom/FXOMDocumentTest.java
```diff
@@ -281,14 +282,61 @@ public void that_missing_imports_during_defines_resolution_cause_exception() thr
         assertEquals(IllegalStateException.class, t.getClass());
         assertTrue(t.getMessage().contains("Bug in FXOMRefresher"));
     }
+
```
jperedadnr, line 285: unneeded extra line

kit/src/test/java/com/oracle/javafx/scenebuilder/kit/fxom/FXOMDocumentTest.java
```diff
@@ -281,14 +282,61 @@ public void that_missing_imports_during_defines_resolution_cause_exception() thr
...
+        
+        ClassLoader classLoader = null;
+        ResourceBundle resources = null;
+        FXOMDocumentSwitch[] switches = new FXOMDocumentSwitch[0]; 
+        
+        Throwable t = assertThrows(ExecutionException.class, 
+                ()->waitFor(()->new FXOMDocument(fxmlText, location, classLoader, resources, switches)));
+        
+        if (t.getCause() != null) {
+            t = t.getCause();
```
jperedadnr, line 301: unused, missing an assertion?

app/src/main/java/com/oracle/javafx/scenebuilder/app/DocumentWindowController.java
```diff
@@ -2355,6 +2362,37 @@ private PreferencesRecordGlobal getPreferencesRecordGlobal() {
...
+        errorDialog.setDetailsTitle(I18N.getString("alert.open.failure.unresolved.imports",
+                                    Integer.toString(missingTypes.size())));
+
+        String allMissing = missingTypes.stream()
+                                        .collect(Collectors.joining(";"+System.lineSeparator()));
+
+        errorDialog.setDebugInfo(I18N.getString("alert.open.failure.unresolved.imports.advice",
+                                 allMissing, missingTypes.size()));
+        errorDialog.setTitle(I18N.getString("alert.title.open") + ": " + fxmlFile.getName());
+        errorDialog.showAndWait();
```
jperedadnr, line 2394: Suggestion: add a button to the dialog with an `Open FXML/JAR Manager` option, so the user can add the missing dependency.

jperedadnr (commented):
Looks really good! Great contribution.
I've left a few comments.

Note also that you might need to disable the `Preview` option if there are unresolved imports, otherwise it fails with a RTE: 
```
java.lang.RuntimeException: Bug in PreviewWindowController::requestUpdate
        at com.gluonhq.scenebuilder.kit@24.0.0-SNAPSHOT/com.oracle.javafx.scenebuilder.kit.preview.PreviewWindowController$1.lambda$run$1(PreviewWindowController.java:269)
  ```

**Comment by Oliver-Loeffler** [comment]
Thank you for the feedback @jperedadnr, I'll look into and rework accordingly.

**Comment by Oliver-Loeffler** [comment]
> Looks really good! Great contribution. I've left a few comments.
> 
> Note also that you might need to disable the `Preview` option if there are unresolved imports, otherwise it fails with a RTE:
> 
> ```
> java.lang.RuntimeException: Bug in PreviewWindowController::requestUpdate
>         at com.gluonhq.scenebuilder.kit@24.0.0-SNAPSHOT/com.oracle.javafx.scenebuilder.kit.preview.PreviewWindowController$1.lambda$run$1(PreviewWindowController.java:269)
> ```

Thanks for the hint. Well, I've reworked it in such a way, that the preview remains working. The unresolvable imports are just removed in process of preview generation.

File: kit/src/main/java/com/oracle/javafx/scenebuilder/kit/fxom/FXOMDocument.java [hunk 5]
```diff
@@ -506,6 +544,44 @@ public enum FXOMDocumentSwitch {
...
+         * switch.
+         * 
+         * @param existingSwitches Array (usually var args) with existing values.
+         * @param newSwitch        New option to be added to the whole set.
+         * @return Array of unique FXOMDocumentSwitch items consisting of the given
+         *         existing switches and the new one.
+         */
+        public static FXOMDocumentSwitch[] combined(FXOMDocumentSwitch[] existingSwitches,
+                                                    FXOMDocumentSwitch newSwitch) {
+            Set<FXOMDocumentSwitch> options = EnumSet.of(FXOMDocumentSwitch.NORMALIZED);
```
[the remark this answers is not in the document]
Oliver-Loeffler: You are right, it is supposed to be `Set<FXOMDocumentSwitch> options = EnumSet.noneOf(FXOMDocumentSwitch.class);`.

kit/src/main/java/com/oracle/javafx/scenebuilder/kit/fxom/FXOMDocument.java
```diff
@@ -102,12 +106,13 @@ public class FXOMDocument {
...
      */
-    public FXOMDocument(String fxmlText, URL location, ClassLoader classLoader, ResourceBundle resources, FXOMDocumentSwitch... switches) throws IOException {
+    public FXOMDocument(String fxmlText, URL location, ClassLoader classLoader, ResourceBundle resources,
+                        FXOMDocumentSwitch... switches) throws IOException {
         this.glue = new GlueDocument(fxmlText);
         this.location = location;
         this.classLoader = classLoader;
         this.resources = resources;
-        initialDeclaredClasses = new ArrayList<>();
+        this.initialDeclaredClasses = new ArrayList<>();
```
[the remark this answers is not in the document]
Oliver-Loeffler: I've used `this` here because it matches the code style used with all the other lines. But yes, its not technically needed.

app/src/main/resources/com/oracle/javafx/scenebuilder/app/i18n/SceneBuilderApp.properties
```diff
@@ -250,6 +250,7 @@ prefs.cssanalyzer.columns.defaults.last = "Defaults" Column Last
 prefs.recent.items = Recent items :
 prefs.animate.accordion = Animate Accordion :
 prefs.wildcard.import = Use Wildcard Imports :
+prefs.preserve.unresolved.import = Load FXML with unresolved imports :
```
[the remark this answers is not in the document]
Oliver-Loeffler: Well, I would not make this change in this PR. I also followed the convention which exists. All other UI labels are formatted in such way, please have a look:

image

Without the space as proposed:
image

This would look strange. If we want to remove the space between text and colon, we should do it for all labels.
image

kit/src/test/java/com/oracle/javafx/scenebuilder/kit/fxom/FXOMDocumentTest.java
Oliver-Loeffler: Indeed, I've now added the 2 missing assertions. The idea was a) to proof a class was not found and b) which one.

```java
        while (t.getCause() != null) {
            t = t.getCause();
        }
        
        assertTrue(t instanceof java.lang.ClassNotFoundException);
        assertEquals("another.unresolvable.Dependency", t.getMessage());
```

app/src/main/java/com/oracle/javafx/scenebuilder/app/DocumentWindowController.java
Oliver-Loeffler: I have no good idea from start where to place this. Would it be okay to consider this as an additional issue and PR?

Oliver-Loeffler: I'll review and make a proposal with a feature request after this has been merged.

pom.xml
```diff
@@ -138,7 +138,7 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-surefire-plugin</artifactId>
-                    <version>3.0.0-M5</version>
+                    <version>3.5.0</version>
```
Maran23: The newest version right now is 3.5.4. Might be worth to update

app/src/main/java/com/oracle/javafx/scenebuilder/app/DocumentWindowController.java
```diff
@@ -2362,6 +2369,37 @@ private PreferencesRecordGlobal getPreferencesRecordGlobal() {
...
+    }
+    
+    private void notifyUserAboutUnresolvableImports(File fxmlFile, List<String> missingTypes, Stage owner) {
+        LOGGER.log(Level.WARNING, "Detected unresolved types in FXML: {0}: {1}",
+                               new Object[] {fxmlFile, missingTypes.stream().collect(Collectors.joining(","))});
+
+        final ErrorDialog errorDialog = new ErrorDialog(owner);
+        String first10 = missingTypes.stream()
+                                     .limit(10)
+                                     .collect(Collectors.joining(";"+System.lineSeparator()));
```
Maran23: missing space here inbetween the `+`

Maran23 (commented):
I think this is a nice feature! Left some comments

app/src/main/resources/com/oracle/javafx/scenebuilder/app/preferences/Preferences.fxml
```diff
@@ -50,7 +50,7 @@
 <?import javafx.scene.layout.RowConstraints?>
 <?import javafx.scene.shape.Rectangle?>
 
-<GridPane hgap="6.0" maxWidth="1.7976931348623157E308" styleClass="theme-presets" vgap="10.0" xmlns="http://javafx.com/javafx/17" xmlns:fx="http://javafx.com/fxml/1">
+<GridPane hgap="6.0" maxWidth="1.7976931348623157E308" prefHeight="-Infinity" prefWidth="550.0" styleClass="theme-presets" vgap="10.0" xmlns="http://javafx.com/javafx/19" xmlns:fx="http://javafx.com/fxml/1">
```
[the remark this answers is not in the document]
Oliver-Loeffler: We could set this in the dialog directly. Whats the reasoning to not have those settings in the FXML?

pom.xml
Oliver-Loeffler: I've updated the plugin version accordingly.

app/src/main/resources/com/oracle/javafx/scenebuilder/app/preferences/Preferences.fxml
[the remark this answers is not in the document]
Oliver-Loeffler: As of now, both values are left overs from editing the FXML file with SceneBuilder at time of development. I've tested it now without both settings and the dialog works fine. Hence I've remove them.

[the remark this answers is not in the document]
Maran23: > We could set this in the dialog directly. Whats the reasoning to not have those settings in the FXML? I am just asking out of curiosity.

It gives us a little more flexibility if we want to reuse the FXML file. 
Lets say we want to use it in a `Dialog`, and also somewhere else, in a normal (non dialog) UI. Then we probably don't want the `prefWidth` for the normal UI, only for the `Dialog`. 
So really really a minor thing! 😄

Maran23:
Tested with many FXMLs, removed some Imports and worked pretty good. The error dialog does not appear when I drag and drop a FXML file, but that might be expected?

Also one interesting note:

When I drag in a FXML with a custom component that is not know by the Scenebuilder, it will not load (as before). 
But: If I remove the custom component import, it will load the FXML, which is a bit odd 😄 but probably expected, since this is really about the missing imports, I guess?

app/src/main/java/com/oracle/javafx/scenebuilder/app/DocumentWindowController.java
```diff
@@ -2362,6 +2369,37 @@ private PreferencesRecordGlobal getPreferencesRecordGlobal() {
...
+    public void showMissingTypesNotificationIfNeeded(File fxmlFile) {
+        FXOMDocument doc = getEditorController().getFxomDocument();
+        if (doc.hasUnresolvableTypes()) {
+            notifyUserAboutUnresolvableImports(fxmlFile, doc.getUnresolvableTypes(), getStage());
+        }
+    }
+    
+    private void notifyUserAboutUnresolvableImports(File fxmlFile, List<String> missingTypes, Stage owner) {
+        LOGGER.log(Level.WARNING, "Detected unresolved types in FXML: {0}: {1}",
+                               new Object[] {fxmlFile, missingTypes.stream().collect(Collectors.joining(","))});
```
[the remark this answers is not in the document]
Oliver-Loeffler: This indeed reads much nicer.

Maran23 (approved):
this looks good to me! Tested various scenarios and they all work for me.

PR #585, opened 2022-10-10
fix: SceneBuilder will not load missing files from recent projects.

### LINKED ISSUE #582: Scene Builder closes after attempt to open a no longer existing file from welcome page recent projects
### Expected Behavior
- Scene Builder starts and presents welcome page
- User selects a recent project / recent file
- Scene Builder detects that this file is missing
- Scene Builder asks user to keep the file in list or to remove the recent file from list
- Scene Builder returns to Welcome Page after user interaction

### Current Behavior
- Scene Builder starts and presents welcome page
- User selects a recent project / recent file
- Scene Builder detects that this file is missing
- Scene Builder shows a notification dialog showing the exception stack trace
- Scene Builder application closes

### Steps to Reproduce
1. Create a project in Scene Builder
2. Save the project to a FXML file
3. Close Scene Builder
4. Delete the FXML file
5. Open Scene Builder
6. Attempt to select the no longer existing FXML file from recent projects list

### Your Environment
- macOS Monterey 12.0.1 (Intel)
- Scene Builder 19 (most likely also 18)

### Screenshots

https://user-images.githubusercontent.com/22102800/194703132-58ec7045-f0f3-4823-825a-5518313cff2d.mov

Oliver-Loeffler:
Scene Builder will ask to remove missing files from recent files list. 
Welcome Page will not close, there will not be an error message.

AskUserToRemoveMissingFile

### Issue

Fixes #582

### Progress


Oliver-Loeffler:
https://user-images.githubusercontent.com/22102800/194958675-a96fac1a-88f9-4c0f-b90b-588389bae6a3.mov

Window positioning is not yet okay. I am hesitating to touch the `AlertDialog` for this. Something like `positionCentralOver` or so could be helpful.

abhinayagarwal:
The new tests added to the app project [doesn't seem to run](https://github.com/gluonhq/scenebuilder/actions/runs/3414142510/jobs/5686120224#step:5:515).

Oliver-Loeffler:
I've moved the surefire-plugin from `kit/POM.xml` to `POM.xml` so that now tests of `app` and `kit` are executed during Maven build.

Oliver-Loeffler:
This PR now additionally addresses #600.

Oliver-Loeffler:
Hi @abhinayagarwal, I have now corrected the POM for app. Tests are running now. Could you please review this one again?

Oliver-Loeffler:
@AlmasB Hi Almas, may be you also could have a look on this one!

Oliver-Loeffler:
Happy new year SceneBuilder!
Hope you'll had a good start into 2023.

@abhinayagarwal, could you please review this one again? The tests are now triggerd properly.

abhinayagarwal:
A very happy new year to you too!

Hopefully, I can get back to this PR some time this week.

Oliver-Loeffler:
I've now merged the latest changes into this one. Also applied reformatting according what checkstyle finds for the change files in main.

abhinayagarwal:
Testing the changes, it looks good on Mac (including the positioning). However, I find the text on the buttons to be too descriptive for a dialog.

Here is a screenshot from IntelliJ IDEA of a dialog when project is not found:

Screenshot 2023-03-30 at 8 52 34 AM

Oliver-Loeffler:
Thanks, very good example. Well I will rework it accordingly tonight.

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
```diff
@@ -496,8 +498,25 @@ private void createEmptyDocumentWindow() {
...
 
+    /**
+     * By default all necessary actions to open a single file or a group of files take place in this method.
+     * If it is required to perform certain actions after successfully loading all files, please use {@code handleOpenFilesAction(List<String> files, Runnable onSuccess)} instead.
+     * 
+     * All error handling takes place here within, there is no way yet to access exceptional results and to work with them.
+     */
     @Override
     public void handleOpenFilesAction(List<String> files) {
+        handleOpenFilesAction(files, null);
```
AlmasB, line 509: Please can we avoid passing `null` into methods (here and in the entire codebase, unless we are forced to by the API) and treat any method parameter as non-nullable by default. So here we could pass `() -> {}`.

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
```diff
@@ -510,21 +529,76 @@ public void handleOpenFilesAction(List<String> files) {
...
+     * @param result    {@link FileOpenResult} holding the list of files to be
+     *                  opened and a map of files with exceptions in case of an
+     *                  error.
+     * @param onSuccess A runnable holding the action to be performed on success.
+     */
+    private void handleFileOpenResult(FileOpenResult result, Runnable onSuccess) {
+        if (result.hasErrors()) {
+            showFileOpenErrors(result);
+        } else {
+            if (onSuccess != null) {
```
AlmasB, line 562: As per above comment, if we treat all parameters as non-null, then the null check is not needed.

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
```diff
@@ -510,21 +529,76 @@ public void handleOpenFilesAction(List<String> files) {
...
+
+    /**
+     * In case of errors (>= 1 exception) a message dialog is shown to the user.
+     * There are versions for exactly 1 error and for more than 1 errors.
+     * 
+     * @param openResult {@link FileOpenResult} holding a map of files with
+     *                   exceptions.
+     */
+    private void showFileOpenErrors(FileOpenResult openResult) {
+        if (openResult.errors.size() == 1) {
```
AlmasB, line 576: I'm not sure we get anything useful out of treating these as separate cases. Since we are making changes here, I think we can come up with a single general case for these, regardless of the number of errors. Maybe something along the lines of:

```
The following file(s) failed to open:
- file 1
- file 2
etc.
```

app/src/main/java/com/oracle/javafx/scenebuilder/app/welcomedialog/WelcomeDialogWindowController.java
```diff
@@ -215,17 +225,105 @@ private void openDocument() {
...
+    protected static AlertDialog questionMissingFilesCleanup(Stage stage, List<String> missingFiles) {
+        String withPath = missingFiles.stream()
+                                      .collect(Collectors.joining(System.lineSeparator()));
+        AlertDialog question = new AlertDialog(stage);
+        question.setDefaultButtonID(ButtonID.CANCEL);
+        question.setShowDefaultButton(true);
+        question.setOKButtonTitle(I18N.getString("alert.welcome.file.not.found.okay"));
+        StringBuilder shortMessage = new StringBuilder();
+        if (missingFiles.size() > 1) {
+            shortMessage.append(I18N.getString("alert.welcome.files.not.found.question"));
```
AlmasB, line 238: Same as in the other error dialog. We should be able to make this a single general case dialog, e.g. the following file(s) are missing...

app/src/main/java/com/oracle/javafx/scenebuilder/app/welcomedialog/WelcomeDialogWindowController.java
```diff
@@ -215,17 +225,105 @@ private void openDocument() {
...
+                    Consumer<List<String>> fileLoader) {
+
+        if (filePaths.isEmpty()) {
+            return;
         }
+
+        var candidates = filePaths.stream()
+                                  .collect(Collectors.groupingBy(this::filePathExists));
+
+        List<String> missingFiles = candidates.getOrDefault(Boolean.FALSE, new ArrayList<>());
```
AlmasB, line 310: All of this is fine, but a bit hard to read the intent. The simplest form is effectively:

```
files.forEach(file -> {
    if (file.exists()) {
        open(file);
    } else {
        showError(file);
    }
});
```

Is there a way we can simplify the code to this form?

AlmasB (commented):
Thanks for this, please see attached comments. Happy to discuss if unclear

abhinayagarwal:
The text on the buttons can be reduced to `Keep missing file(s)` and `Remove missing files(s)`, as the user clicked on it and already knows that they belong to "recent projects". The Alert additional mentions the same in its description.

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
Oliver-Loeffler: Agree, thats now implemented. Very good proposal, thank.

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
Oliver-Loeffler: Agree, thats now implemented. Very good proposal, thank.

app/src/main/java/com/oracle/javafx/scenebuilder/app/welcomedialog/WelcomeDialogWindowController.java
Oliver-Loeffler: I've reworked the code accordingly.

app/src/main/java/com/oracle/javafx/scenebuilder/app/welcomedialog/WelcomeDialogWindowController.java
Oliver-Loeffler: Thanks, I agree with that and will simplify the dialog accordingly.

Oliver-Loeffler: I've remove the condition and the customized dialog. There is now one dialog for n=1 and n>1.

Oliver-Loeffler:
@abhinayagarwal I've updated the dialog as proposed by you and @AlmasB.

Oliver-Loeffler:
Interestingly this one no longer works with OpenJFX21 ... I'm looking into it.

Oliver-Loeffler:
The reworked dialog looks now as follows:

ReworkedDialog

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
Oliver-Loeffler: Well, indeed for this PR it is not needed as SceneBuilder effectively cannot open multiple files. But there is another PR which enables Drag&Drop and there this becomes beneficial as then one can skim through the error messages.

abhinayagarwal:
image

Can the buttons be right aligned?

abhinayagarwal:
> Besides the "remove missing file" dialog, this version also shows a dialog with the stack trace for errors occuring during FXML loading. This dialog is shown when load is attempted from welcome dialog or from file menu.

I am wondering if it really makes sense to show the user of the application the stack trace 🤔

Oliver-Loeffler:
> image
> 
> Can the buttons be right aligned?

I'll see what I can do.

Oliver-Loeffler:
Well, the Windows version of the dialog is intentionally designed the way, that the buttons are centered. The Mac dialog looks different. Actually, if we also want to change the dialog appearance, this would be a different PR. So, I would like to keep the appearance as it is. Its not only this dialog, it would impact all Windows ErrorDialogs (AlertDialogs).

Oliver-Loeffler:
Centered, on Windows, this would look like:
Centered

Oliver-Loeffler:
With the reworked ErrorDialog it is also now possible to have a useful title for the error detail. Also in SceneBuilderApp the dialog is now called in such way, so we always have a correct modality and one cannot loose a window.

As mentioned before, the error detail is a detail, the dialog requires user action to open.

ModalityEditor

ModalityWelcome

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
```diff
@@ -137,7 +140,7 @@ public static SceneBuilderApp getSingleton() {
     public SceneBuilderApp() {
         assert singleton == null;
         singleton = this;
-
+
```
abhinayagarwal, line 143: please remove this

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
```diff
@@ -507,24 +527,81 @@ public void handleOpenFilesAction(List<String> files) {
         }
 
         EditorController.updateNextInitialDirectory(fileObjs.get(0));
-
+
```
abhinayagarwal, line 530: same here

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
```diff
@@ -507,24 +527,81 @@ public void handleOpenFilesAction(List<String> files) {
         }
 
         EditorController.updateNextInitialDirectory(fileObjs.get(0));
-
+        
         // Fix for #45
         if (userLibrary.isFirstExplorationCompleted()) {
-            performOpenFiles(fileObjs);
+            var openResult = performOpenFiles(fileObjs);
+            handleFileOpenResult(openResult, onSuccess, WelcomeDialogWindowController.getInstance().getStage());
```
abhinayagarwal, line 534: These 2 steps are dependent on each other. I am wondering if we should merge them. This way we will be able to do:

```
performOpenFiles(fileObjs, onSuccess);
```

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
Oliver-Loeffler: Was removed.

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
Oliver-Loeffler: Rmoved.

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
Oliver-Loeffler: Good catch. There is now a new method combining this.

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
```diff
@@ -656,12 +746,15 @@ public DocumentWindowController getFrontDocumentWindow() {
         return null;
     }
 
-    private void performOpenFiles(List<File> fxmlFiles) {
+    private FileOpenResult performOpenFiles(List<File> fxmlFiles) {
```
abhinayagarwal, line 749: My last comment to update `performOpenFiles` was to update this method instead. Since `handleFileOpenResult` only does the following:

```
        if (result.hasErrors()) {
            showFileOpenErrors(result, owner);
        } else {
            onSuccess.run();
        }
```

it should be possible to incorporate it in the same method by changing it to `void performOpenFiles(final List<File> files, Runnable onSuccess)`. This way we won't need `FileOpenResult` record as well.

Or, do you see a potential upside of having the record and the additional methods to handle everything?

Oliver-Loeffler: The potential I see is separation of concerns and with that the potential to get this all testable.
The reason why I chose this way was to separate UI actions, file handling and error handling, so that we might later implement tests. Also (for later) when drag&drop can be used, we would have a result type which bundles the file to be opened with a potential exception which occured.

This way one could even move the performOpenFiles logic into a separate class and make the SceneBuilderApp class a little smaller. That would move the design more towards building blocks.

Oliver-Loeffler: I'll try to consolidate the behavior and implement a test for the file open logic accordingly.

abhinayagarwal: > The reason why I chose this way was to separate UI actions, file handling and error handling, so that we might later implement tests.

We can file a separate issue and work on it in a PR of its own.

Oliver-Loeffler: I agree with that approach.

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
```diff
@@ -83,6 +83,8 @@
 import java.nio.file.Paths;
 import java.time.LocalDate;
 import java.util.*;
+import java.util.function.*;
```
abhinayagarwal, line 86: avoid wildcard import

app/src/main/resources/com/oracle/javafx/scenebuilder/app/i18n/SceneBuilderApp_ja.properties
```diff
@@ -394,10 +394,6 @@ alert.title.messagebox = 外部プログラムで開く
 
 alert.open.failure1.message = ''{0}''を開けませんでした
 alert.open.failure1.details = 開くのに失敗しました。選択されたファイルが有効なFXMLドキュメントであることを確認してください。
-alert.open.failureN.message = 指定されたファイルを開けませんでした
-alert.open.failureN.details = 開くのに失敗しました。それらのファイルが有効なFXMLドキュメントであることを確認してください。
-alert.open.failureMofN.message = {0}個のファイルを開けませんでした
-alert.open.failureMofN.details = いくつかのファイルを開くのに失敗しました。それらのファイルが有効なFXMLドキュメントであることを確認してください。
```
abhinayagarwal, line 400: with https://github.com/gluonhq/scenebuilder/pull/683 now merged, we have to do this for the chinese translation file as well.

app/src/test/java/com/oracle/javafx/scenebuilder/app/JfxInitializer.java
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright (c) 2021, 2024, Gluon and/or its affiliates.
```
abhinayagarwal, line 2: new files only need the year of initialization on them:

```
Copyright (c) 2024, Gluon and/or its affiliates.
```

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
Oliver-Loeffler: done

Oliver-Loeffler:
If this PR gets approved and merged, I'd like to continue with #576 which then would fix #120.

abhinayagarwal (approved):
LGTM!

Oliver-Loeffler:
It took a while but actually I am also now happy with the outcome. Thanks for your feedback.

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
```diff
@@ -507,24 +533,60 @@ public void handleOpenFilesAction(List<String> files) {
...
         if (userLibrary.isFirstExplorationCompleted()) {
-            performOpenFiles(fileObjs);
+            performOpenFiles(fileObjs, onError, onSuccess);
         } else {
             // open files only after the first exploration has finished
             userLibrary.firstExplorationCompletedProperty().addListener(new InvalidationListener() {
                 @Override
                 public void invalidated(Observable observable) {
                     if (userLibrary.isFirstExplorationCompleted()) {
-                        performOpenFiles(fileObjs);
```
AlmasB, line 520: Previously `performOpenFiles()` was called before removing the user library listener, now it's after. Does the order make any difference?

AlmasB (approved):

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
Oliver-Loeffler: I have to admit that I was thinking about this as well, but eventually I found no difference. I kind of dislike how this is handled but it works. Cleaning this up could be a separate PR. At least I have not observed a difference in behavior. Hence I've kept it this way. The fileObjs list is not an observable one or so, its just plain data with no other things attached so I assumed this as saf.e

kit/src/main/java/com/oracle/javafx/scenebuilder/kit/editor/panel/util/dialog/ErrorDialog.java
```diff
@@ -43,6 +43,7 @@
 public class ErrorDialog extends AlertDialog {
     
     private String debugInfo;
+    private String detailsTitle;
```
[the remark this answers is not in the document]
Oliver-Loeffler: Good point.  But, even if there would be a default title, one would need to prevent that null is assigned as a new value. So the null check would effectively move into the setter. But indeed , there is no null check for the debugInfo. Nevertheless, I'd consider such things as a separate PR. Thanks for the review!

Oliver-Loeffler (commented):
Thanks for the feedback @abhinayagarwal and @AlmasB. That seems to be okay now. The issues with the error dialog are valid but should be resolved in a separate PR.

kit/src/main/java/com/oracle/javafx/scenebuilder/kit/editor/panel/util/dialog/ErrorDialog.java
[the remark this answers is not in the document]
AlmasB: > I'd consider such things as a separate PR

Agreed.

> one would need to prevent that null is assigned as a new value

If someone is passing null via public API, that's their problem 😄

Oliver-Loeffler:
This one fixed also #600.

PR #589, opened 2022-10-19
fix: Zoom-in/out using keyboard shortcut not working correctly

### LINKED ISSUE #424: Zoom-in/out using keyboard shortcut not working correctly
SceneBuilder offers key bindings for zooming in `[CTRL]`+`[+]` and out `[CTRL]`+`[/]` in FXML preview.
These are not working on Windows 10 for Scene Builder 17.

### Expected Behavior
While haveing no active selections in Scene Builder editor, pressing `[CTRL]`+`[+]` should result in zoom-in by pre-defined step and pressing `[CTRL]`+`[/]` should result in zooming out.

### Current Behavior

On Windows:
Currently no reaction is visible while pressing `[CTRL]`+`[+]` or `[CTRL]`+`[/]`. Zoom works only by clicking the corresponding menu items.

On Linux (Ubuntu 20.04.3 LTS)
Zoom-in using `[CTRL]`+`[+]` works when the `[+]` key on the alphabetic part of the keyboard is used. The `[+]` key on numpad does does not cause any reaction. Zoom-out using `[CTRL]`+`[/]` is not working at all.

### Steps to Reproduce
* Install Scene Builder 17 in Windows 10
* Create a new FXML with any controls inside or open an existing FXML file
* Attempt to change zoom level via `[CTRL]`+`[+]` or `[CTRL]`+`[/]`
* Ensure that german keyboard layout is used and numpad is active

### Your Environment
* Windows 10 Pro (Version: 21H1, 19043.1348)
* Ubuntu 20.04.3 LTS
* Both with german keyboard layout and active numpad (numlock enabled).

### Screenshots

20211125_SceneBuilder_Zoom_KeyBindingsNotWorking

Oliver-Loeffler:
Zoom-In/Out was supposed to work with keyboard shortcuts.
For Zoom-In `MODIFIER`+`+` and for Zoom-Out `MODIFIER`+`/` were defined.
For some reason the predefined KeyCombinations were not working.
Also, many keyboards have two keys for `+` (the PLUS key and the ADD key) and the same applies to `/`, which is mapped to SLASH and DIVIDE.

The `DocumentWindowController` offers a `mainKeyEventFilter` where those double accelerators can be tracked.
The class `MenuBarController` now handles the mappings to ADD and DIVIDE whereas PLUS and SLASH are handled by the `DocumentWindowController`.

### Issue

Fixes #424

### Progress

### DIFF
--- a/app/src/main/java/com/oracle/javafx/scenebuilder/app/menubar/MenuBarController.java
+++ b/app/src/main/java/com/oracle/javafx/scenebuilder/app/menubar/MenuBarController.java
@@ -209,7 +209,11 @@ public class MenuBarController {
     private MenuItem showSampleControllerMenuItem;
     @FXML
     private Menu zoomMenu;
-
+    
+    private ZoomInActionController zoomInController;
+    
+    private ZoomOutActionController zoomOutController;
+    
     // Modify
     @FXML
     private MenuItem fitToParentMenuItem;
@@ -1269,17 +1273,19 @@ public class MenuBarController {
      */
     
     final static double[] scalingTable = {0.25, 0.50, 0.75, 1.00, 1.50, 2.0, 4.0};
-    
+
     private void updateZoomMenu() {
         final double[] scalingTable = {0.25, 0.50, 0.75, 1.00, 1.50, 2.0, 4.0};
 
+        zoomInController = new ZoomInActionController();
         final MenuItem zoomInMenuItem = new MenuItem(I18N.getString("menu.title.zoom.in"));
-        zoomInMenuItem.setUserData(new ZoomInActionController());
+        zoomInMenuItem.setUserData(zoomInController);
         zoomInMenuItem.setAccelerator(new KeyCodeCombination(KeyCode.ADD, modifier)); //NOI18N
         zoomMenu.getItems().add(zoomInMenuItem);
 
+        zoomOutController = new ZoomOutActionController();
         final MenuItem zoomOutMenuItem = new MenuItem(I18N.getString("menu.title.zoom.out"));
-        zoomOutMenuItem.setUserData(new ZoomOutActionController());
+        zoomOutMenuItem.setUserData(zoomOutController);
         zoomOutMenuItem.setAccelerator(new KeyCodeCombination(KeyCode.DIVIDE, modifier));  //NOI18N
         zoomMenu.getItems().add(zoomOutMenuItem);
         
@@ -1294,7 +1300,20 @@ public class MenuBarController {
         }
     }
 
+    public void zoomIn() {
+        runActionController(zoomInController);
+    }
     
+    public void zoomOut() {
+        runActionController(zoomOutController);
+    }
+    
+    private void runActionController(MenuItemController controllerToRun) {
+        if (controllerToRun.canPerform()) {
+            controllerToRun.perform();
+        }
+    }
+
     private static int findZoomScaleIndex(double zoomScale) {
         int result = -1;
         
--- a/app/src/main/java/com/oracle/javafx/scenebuilder/app/menubar/MenuBarController.java
+++ b/app/src/main/java/com/oracle/javafx/scenebuilder/app/menubar/MenuBarController.java
@@ -1275,12 +1275,12 @@ public class MenuBarController {
 
         final MenuItem zoomInMenuItem = new MenuItem(I18N.getString("menu.title.zoom.in"));
         zoomInMenuItem.setUserData(new ZoomInActionController());
-        zoomInMenuItem.setAccelerator(new KeyCharacterCombination("+", modifier)); //NOI18N
+        zoomInMenuItem.setAccelerator(new KeyCodeCombination(KeyCode.ADD, modifier)); //NOI18N
         zoomMenu.getItems().add(zoomInMenuItem);
-        
+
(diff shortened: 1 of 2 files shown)


abhinayagarwal:
Testing on Mac

### Preview app

both CMD + = and CMD + SHIFT + = zooms in
both CMD + - and CMD + SHIFT + - zooms out

### SceneBuilder app

CMD + = does nothing. 
CMD + SHIFT + = zooms in
CMD + - zooms out
CMD + SHIFT + - zooms all out at once

Oliver-Loeffler:
I will review this and complement the table above with our findings. Actually I've tested the above table on an older x86 Macbook with MacOS 10.12.

app/src/main/java/com/oracle/javafx/scenebuilder/app/menubar/MenuBarController.java
```diff
@@ -1269,22 +1274,36 @@ private void handleOnActionMenu(MenuItem i) {
...
-        zoomOutMenuItem.setUserData(new ZoomOutActionController());
-        zoomOutMenuItem.setAccelerator(new KeyCharacterCombination("/", modifier));  //NOI18N
+        zoomOutMenuItem.setUserData(zoomOutController);
         zoomMenu.getItems().add(zoomOutMenuItem);
-        
+        if (EditorPlatform.IS_MAC) {
+            zoomInMenuItem.setAccelerator(new KeyCharacterCombination("+", modifier)); // NOI18N
+            zoomOutMenuItem.setAccelerator(new KeyCharacterCombination("-", modifier)); // NOI18N
+            // Neither KeyCode.MINUS or KeyCode.SUBTRACT seem to work on macOS.
+            // No chance to test on keyboard with num key pad yet.
```
abhinayagarwal, line 1294: This is definitely a strange behavior 🤔

abhinayagarwal: Yes, I can reproduce this on mac.

abhinayagarwal:
@Oliver-Loeffler If you need any help, feel free to ping me. It would be nice to have this PR merged before we merge #619 and release SB 20.

Oliver-Loeffler:
I will look into this one tonight.

Oliver-Loeffler:
Well I just learned that Zooming was also a thing for the Preview. Never thought about this one, just worked on the Zoom function for the editor window. Indeed, I'll update this accordingly.

I have tested the key strokes as described in the table above. 
With the keys you have posted, I have no zooming at all for preview.

Testing on Mac OS 12.0.1
Preview app

both CMD + = and CMD + SHIFT + = zooms in    -> cannot reproduce
both CMD + - and CMD + SHIFT + - zooms out   -> cannot reproduce

SceneBuilder app

CMD + = does nothing.    -> I have no `key` -> cannot reproduce
CMD + SHIFT + = zooms in   -> does nothing
CMD + - zooms out   -> yes, confirm, can reproduce
CMD + SHIFT + - zooms all out at once -> yes, confirm, can reproduce

I'll take an image of my keyboard. Unfortunately I have no numpad for my Mac.

app/src/main/java/com/oracle/javafx/scenebuilder/app/menubar/MenuBarController.java
Oliver-Loeffler: May be this requires a different approach depending on the keyboard type. Are you using a separate keyboard or numpad? I've just an old laptop available.

app/src/main/java/com/oracle/javafx/scenebuilder/app/menubar/MenuBarController.java
```diff
@@ -1299,7 +1318,61 @@ private void updateZoomMenu() {
...
+        // is not properly accepted hence this handling here.
+        if (EditorPlatform.IS_MAC && "-".equals(event.getText())) {
+            runActionController(zoomOutController, event);
+            return;
+        }
+
+        if (EditorPlatform.IS_MAC && "+".equals(event.getText())) {
+            runActionController(zoomInController, event);
+            return;
+        }
```
Oliver-Loeffler, line 1348: There is also now special handling of `CMD`+`+` for MacOS in the MenuBar controller. Could you please try this @abhinayagarwal ?

Oliver-Loeffler:
I will not get this one completed before end of next week. As I started digging into this topic I found, that the proposed key combinations using `+` or `-` (or even `=`) are not available for many keyboard layouts. My proposal here is to use more generic combinations which are available on most layouts, such as `CMD`+`up` for zoom-in and `CMD`+`down` for zoom-out. I've also tested this PR using an external Apple keyboard with numpad and it works as expected - given one has a german key layout. This is the point, pre-configured keys for zoom will break or work depending on the keyboard layout used. 

Eventually its even thinkable to make those key combinations configurable. As of now, I have not yet found how to detect which keyboard layout is effectively used for a system. If this would be detectable, one could generate key combinations for certain layouts and a generic combination which fits all.

On the other hand, it seems that zoom is actually not yet implemented for the preview window. If this works in some cases, then by accident. This should be solved by a separate PR.

What do you think about this @abhinayagarwal ?

jperedadnr (commented):
Tested on macOS, and it looks good. I've left a few comments.

PR #593, opened 2022-10-20
fix: Prevents CTRL+P from being masked by a given accelerator defined in FXML (#553)

### LINKED ISSUE #553: Version 18.0.0 bug.
Product Version
JavaFX Scene Builder 18.0.0

Build Information
Version 18.0.0
Date: 2022-03-31 20:20:30
JavaFX Version: 18
Java Version: 17, OpenJDK Runtime Environment

### Expected Behavior

Ctrl+P should invoke menu command "Show Preview in Window".

### Current Behavior

Ctrl+P does nothing.

### Steps to Reproduce

Invoke Ctrl+P.

### Your Environment

Windows 10.

Oliver-Loeffler:
This quick fix will prevent `CTRL`+`P` from being blocked by a accelerator defined in an FXML. `CTRL`+`P` is used to display the preview window and it works with some FXMLs, not with others.

A clean solution to this requires lot more work. With the current way the Scene Builder works, the problem can occurr with any other accelerator defined in Scene Builder.

### Issue

Fixes #553

### Progress


app/src/main/java/com/oracle/javafx/scenebuilder/app/DocumentWindowController.java
```diff
@@ -316,6 +316,13 @@ public enum ActionStatus {
             }
             event.consume();
         }
+        
+        // Keep preview function working even if preview accelerator is 
+        // already occupied in FXML
+        if ("p".equalsIgnoreCase(event.getText()) && modifierDown) {
```
abhinayagarwal, line 322: Can we replace this with `KeyCode.P.equals(event.getCode()) && modifierDown` ?

Oliver-Loeffler: Yes, I have replaced this expression with the proposal. Also tested it on macOS 12, Fedora 36 and Win10 and it works on all 3 platforms.

jperedadnr (approved):
This looks good. I've tested with an FXML file that has the accelerator on Mac (but it didn't happen) and Windows (failed before this PR, passes after it).

I wonder why the accelerator from the FXML gets installed and somehow takes precedence, but your fix won't hurt in any case.

abhinayagarwal (approved):

Oliver-Loeffler:
I actually have to admit, that this behavior also irritated me. However, it seems to be confusing. I'm updating this branch to see if it still works after all those recent changes.

PR #619, opened 2023-03-23
build: Use Java and JavaFX 20

### LINKED ISSUE #617: build: Update Scene Builder pom.xml & Actions to make use of JavaFX 20 and Java 20
### Expected Behavior
Scene Builder should support JavaFX 20.
OpenJFX 20 is now available, also latest Java 20.
Hence Scene Builder could be moved to the new platform.

### Current Behavior
Packaged Scene Builder is shipped with JDK 19 based runtime and OpenJFX 19.

### Context
Housekeeping & Maintenance, keep Scene Builder on par with latest platform development.

Oliver-Loeffler:
Updated Github Actions Workflows to use JavaFX 20 and Java 20.

### Issue

Fixes #617

### Progress

The workflows did not run yet, so it can happen that for the one or other workflows errors occur. I will run tests locally on Ubuntu and Windows accordingly, Mac test only for x86.

### DIFF
--- a/kit/src/test/resources/com/oracle/javafx/scenebuilder/kit/fxom/HeaderWithNotOnlyImports.fxml
+++ b/kit/src/test/resources/com/oracle/javafx/scenebuilder/kit/fxom/HeaderWithNotOnlyImports.fxml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
-<?language javascript?>
+<?language nameOfScriptLanguage?>
 
 <!-- This could be a license header -->
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
         uses: oracle-actions/setup-java@v1
         with:
           website: jdk.java.net
-          release: 19
+          release: 20
 
       - name: Run Tests (Linux)
         if: runner.os == 'Linux'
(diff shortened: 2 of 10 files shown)


kit/src/test/resources/com/oracle/javafx/scenebuilder/kit/fxom/HeaderWithNotOnlyImports.fxml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
-<?language javascript?>
+<?language nameOfScriptLanguage?>
```
AlmasB, line 3: I assume this is a required change related to the recent jfx 20 update

AlmasB (approved):
Apart from the 1 comment, the rest looks fine

kit/src/test/resources/com/oracle/javafx/scenebuilder/kit/fxom/HeaderWithNotOnlyImports.fxml
Oliver-Loeffler: Its related to fact that the "FXML JavaScript Engine" is now disabled by default. Hence the related test will fail with an error. However, for the actual test it is not of interest which language is used. Hence for the test its okay to attempt loading the FXML like so. It would have been possible to enable the engine just for this test but this is indeed not needed.

Reference: https://gluonhq.com/products/javafx/openjfx-20-release-notes/

abhinayagarwal (approved):

PR #639, opened 2023-05-11
refactor: Clean up DocumentWatchingController

AlmasB:
General housekeeping:
- Added `Optional<>` return methods to replace commonly used null checks, simplifying existing and future code
- Clean up DocumentWatchingController, no functionality changes except for `IllegalStateException` "bugs", which no longer exist
- Removed ScenicViewStarter, which was effectively an empty file

### Issue

Related to #445 

### Progress


app/src/main/java/com/oracle/javafx/scenebuilder/app/DocumentWatchingController.java
```diff
@@ -58,49 +58,31 @@ public class DocumentWatchingController implements FileWatcher.Delegate {
...
     public DocumentWatchingController(DocumentWindowController documentWindowController) {
         this.documentWindowController = documentWindowController;
         this.editorController = documentWindowController.getEditorController();
         this.resourceController = documentWindowController.getResourceController();
         this.sceneStyleSheetMenuController = documentWindowController.getSceneStyleSheetMenuController();
         
-        this.editorController.sceneStyleSheetProperty().addListener(
-                (ChangeListener<ObservableList<File>>) (ov, t,
-                        t1) -> update());
+        this.editorController.sceneStyleSheetProperty().addListener((ov, t, t1) -> update());
```
abhinayagarwal, line 68: should we replace the change listener with an invalidation listener?

app/src/main/java/com/oracle/javafx/scenebuilder/app/DocumentWatchingController.java
```diff
@@ -58,49 +58,31 @@ public class DocumentWatchingController implements FileWatcher.Delegate {
...
     public void update() {
-        /*
-         * The file watcher associated to this document window controller watches:
-         *  1) the file holding the FXML document (if any)
-         *  2) the resource file set in the Preview menu (if any)
-         *  3) the style sheets files set in the Preview menu (if any)
-         */
-        
-        final List<Path> targets = new ArrayList<>();
+        List<Path> targets = new ArrayList<>();
```
abhinayagarwal, line 80: does removing `final` has any advantage?

Oliver-Loeffler: Well, one could also ask, why declare variables as final? For me, getting rid of superfluous declarations is okay. Well, I personally consider this as superfluous here. Another aspect could be just to avoid reassignemnts later somewhere, this would not work with final.

But eventually, the lesser is written, the easier to read.

app/src/main/java/com/oracle/javafx/scenebuilder/app/DocumentWatchingController.java
AlmasB: Actually, not sure. Is there a semantic difference in this case? This is the sequence I am imagining:

1. the stylesheet property becomes invalid
2. we `update()` (and through `editorController.getSceneStyleSheets()` we effectively ask the property to re-compute itself)

Overall, I think it would behave in a way similar to the change listener (possibly in a different order compared to above). Though I'm not brave enough to make the switch without a proper test 😃

app/src/main/java/com/oracle/javafx/scenebuilder/app/DocumentWatchingController.java
AlmasB: I am not against `final` per se, but as Oliver says, my reasoning was that the scope of `targets` is trivial and the use of `final` is a bit of an overkill.

app/src/main/java/com/oracle/javafx/scenebuilder/app/DocumentWatchingController.java
abhinayagarwal: There shouldn't be any semantic difference. I prefer to avoid change listeners if the `old` or `new` values of the property are not used.

We can keep it as it is for now.

abhinayagarwal (approved):

Oliver-Loeffler (approved):

PR #657, opened 2023-10-03
fix: hsb precision opacity

### LINKED ISSUE #655: Inadequate HSB Precision Causes RGB Values to Change When Adjusting Opacity
**Issue Summary**:
Adjusting the opacity of a selected color in SceneBuilder results in an unintended change in its RGB values. This is likely due to the inadequate precision of the HSB color representation used.

### Expected Behavior

When adjusting the opacity of a selected color, only the alpha (transparency) value should change; the RGB values should remain consistent.

### Current Behavior

After selecting a color and then adjusting its opacity, the RGB values of the color shift. This is probably due to converting the color from an HSB representation with inadequate precision.

### Proposed Solutions
1. Increasing the precision of the HSB color representation by one decimal point can provide accurate results for the majority of cases.
2. If the precision is increased by two decimal points, it will allow for accurate restoration of the vast majority of colors.

### Steps to Reproduce

1. Select a color in the SceneBuilder's ColorPicker.
2. Adjust the opacity of the selected color.
3. Observe a change in the RGB values, even though only the opacity was modified.

### Your Environment

- SceneBuilder version: (8),20,21
- Operating System: Windows 10
- Java version:  (8),20,21

### Screenshots

rbg_changed

leewyatt:
**PR Summary**:
This PR addresses the issue where adjusting the opacity of a selected color in SceneBuilder results in unintended changes in its RGB values. This is due to inadequate precision of the HSB color representation.

### Issue

Fixes #655 

### Progress


abhinayagarwal:
The representation of these properties is in Integer to the users of Scene Builder. I am not sure if converting it to double is the best solution.

leewyatt:
@abhinayagarwal Thank you for your insightful feedback regarding the PR. I understand your concerns about changing the data representation from Integer to Double for the color properties in Scene Builder, and how this might not align with the user experience and the current UI design.

I respect your perspective and agree that maintaining consistency and user friendliness is crucial. I’m open to discussing this further or exploring alternative approaches that could address the opacity adjustment issue without altering the core data types presented to the users.

Given your feedback, I suggest we could close this PR for now. I am more than willing to work on refining the solution based on our discussion and any additional guidance you might have.

PR #658, opened 2023-10-18
fix: logging file creation from jars and package

### LINKED ISSUE #661: Logging is failing
This bug was introduced in 2021 via https://github.com/gluonhq/scenebuilder/pull/371. The approach (setting a file property) did not work, as the JDK's LogManager failed to find the file -- inside a jar -- and silently fails (frustratingly), resulting in no logging config at all.

### Expected Behavior

Log files should be created

### Current Behavior

Log files are not created while running from jar and packages

### Steps to Reproduce

Install and run the latest Scene Builder package

### Your Environment

Version 21.0.0-SNAPSHOT
JavaFX Version: 21
Java Version: 21, OpenJDK Runtime Environment

credmond:
Fixed logging file, which previously only worked from an IDE / local run outside of a JAR. This bug was introduced in 2021; https://github.com/gluonhq/scenebuilder/pull/371. The approach (setting a file property) did not work, as the JDK's LogManager failed to find the file -- inside a jar -- and silently fails (frustratingly), resulting in no logging config at all. That is not how that property is intended to be used (an external file is assumed as FileInputStream is used). Another approach that would work is to use java.util.logging.config.class and specify a class (with the job of setting properties). This fix is a shortcut to basically do the same thing.

### Issue
No logging file was created nor logged to.

Fixes #661 

As per commit message.

### Progress

### DIFF
--- a/app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
+++ b/app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
@@ -34,52 +34,32 @@ package com.oracle.javafx.scenebuilder.app;
 
 import com.oracle.javafx.scenebuilder.app.DocumentWindowController.ActionStatus;
 import com.oracle.javafx.scenebuilder.app.about.AboutWindowController;
-import com.oracle.javafx.scenebuilder.kit.preferences.MavenPreferences;
-import com.oracle.javafx.scenebuilder.kit.ResourceUtils;
-import com.oracle.javafx.scenebuilder.kit.ToolTheme;
-import com.oracle.javafx.scenebuilder.kit.alert.ImportingGluonControlsAlert;
-import com.oracle.javafx.scenebuilder.kit.alert.SBAlert;
 import com.oracle.javafx.scenebuilder.app.i18n.I18N;
 import com.oracle.javafx.scenebuilder.app.menubar.MenuBarController;
 import com.oracle.javafx.scenebuilder.app.preferences.PreferencesController;
 import com.oracle.javafx.scenebuilder.app.preferences.PreferencesRecordGlobal;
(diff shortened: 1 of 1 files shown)


abhinayagarwal:
I have tested the fix against source, jar and dmg package. It runs under all conditions.

abhinayagarwal (dismissed):

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
```diff
@@ -124,7 +119,12 @@ public enum ApplicationControlAction {
     private final BooleanBinding startupTasksFinished = Bindings.isEmpty(startupTasks);
 
     static {
-        System.setProperty("java.util.logging.config.file", SceneBuilderApp.class.getResource("/logging.properties").getPath());
+        try {
+            // Reads logging property file from local runs, or embedded in a JAR
+            LogManager.getLogManager().readConfiguration(SceneBuilderApp.class.getResourceAsStream("/logging.properties"));
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
abhinayagarwal, line 126: Can we add a message in the runtime exception?

> Unable to load configuration file logging.properties

credmond:
> I have tested the fix against source, jar and dmg package. It runs under all conditions.

Thank you!

app/src/main/java/com/oracle/javafx/scenebuilder/app/SceneBuilderApp.java
credmond: Done there now with a second commit.

credmond:
Sorry, I renamed the title after you renamed it (I didn't realise -- I thought it was my original), I put it back to your version... :)

PR #730, opened 2024-09-27
feat: Remove deprecated API from JavaFX 23

### LINKED ISSUE #726: javafx.css.SimpleSelector in javafx.css has been deprecated and marked for removal
`javafx.css.SimpleSelector` and `javafx.css.CompoundSelector` are deprecated in 23, see https://bugs.openjdk.org/browse/JDK-8324182, and removed in 24-ea, see https://bugs.openjdk.org/browse/JDK-8323706.

These are used in Kit `CSSInternal::getStyleClasses`, and the fix should be simple, just using this code:

```
for (Rule r : s.getRules()) {
            for (Selector ss : r.getSelectors()) {
                styleClasses.addAll(ss.getStyleClassNames());
            }
        }
```

jperedadnr:
### Issue

Fixes #726

### Progress

### DIFF
--- a/kit/src/main/java/com/oracle/javafx/scenebuilder/kit/util/CssInternal.java
+++ b/kit/src/main/java/com/oracle/javafx/scenebuilder/kit/util/CssInternal.java
@@ -59,18 +59,16 @@ import com.oracle.javafx.scenebuilder.kit.editor.EditorController;
 import com.oracle.javafx.scenebuilder.kit.editor.EditorPlatform.Theme;
 import com.oracle.javafx.scenebuilder.kit.fxom.FXOMInstance;
 import com.oracle.javafx.scenebuilder.kit.metadata.property.ValuePropertyMetadata;
-import javafx.css.CompoundSelector;
 import javafx.css.Rule;
 import javafx.css.Selector;
-import javafx.css.SimpleSelector;
 import javafx.css.Style;
 import javafx.css.Stylesheet;
 import javafx.css.CssParser;
 
 /**
  *
- * Utility classes using css internal classes (from com.sun package). Note that
- * the CSS Analyzer is also using extensively com.sun classes.
+ * Utility classes to extract CSS information.
+ * Note: Requires JavaFX 23+
  *
  */
 public class CssInternal {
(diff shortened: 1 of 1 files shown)


abhinayagarwal:
Please update the title to follow [pre-defined format](https://github.com/gluonhq/scenebuilder/blob/master/CONTRIBUTING.md)

jperedadnr:
Done, thought I'm not sure "feature" is the right one for this issue.

kit/src/main/java/com/oracle/javafx/scenebuilder/kit/util/CssInternal.java
```diff
@@ -59,18 +59,16 @@
...
 
 /**
  *
- * Utility classes using css internal classes (from com.sun package). Note that
- * the CSS Analyzer is also using extensively com.sun classes.
+ * Utility classes to extract CSS information.
+ * Note: Requires JavaFX 23+
  *
  */
 public class CssInternal {
```
Oliver-Loeffler, line 74: Hi @jperedadnr , for me how SceneBuilder internally handles CSS is at this moment not very transparent. Especially with having the CSS inspector not working (since version 9), it would be good to have basic tests attached which illustrate how Scene Builder handles CSS and how things are supposed to work.

Is it possible that you could add a minimalistic test here with a bried description how things should work?
We could use this as a starting point for an architectural description / documentation.

Oliver-Loeffler (changes requested):
Ideally there is a test, showing that this is working as intended. The test should not reflect the exact class and package structure from Scene Builder kit but it shoud kind of demonstrate how CSS handling works and how the different CSS classes play together. May be a new test package such as `com.oracle.javafx.scenebuilder.kit.tests.css` could be a good starting point.

kit/src/main/java/com/oracle/javafx/scenebuilder/kit/util/CssInternal.java
Oliver-Loeffler: I was looking over it and indeed, its all private API. Thats not the ideal place to start integration of tests. Technically this seems to work. So this is actually okay for me.

Oliver-Loeffler (approved):

PR #739, opened 2024-10-01
fix: Add MSI package for Windows to enable per-machine installs.

### LINKED ISSUE #672: Please provide a system wide installer
Users are not allowed to install and run software from their userprofile in restricted environments.

### Expected Behavior
Installation should offer an installation to %ProgramFiles%  (parameter ALLUSERS=1)

### Current Behavior
Current installer tries to install the Application into the user profile

Oliver-Loeffler:
Scene Builder is published for Windows using a per-user installation MSI package, which is created by JPackage.
In certain scenarios, a per-machine install is desired. 

Unfortunately, when using JPackage, one has to decide upfront between per-user or per-machine install. Having an option in the installer would require an update to JPackage or to the WXS/WXI files used by JPackage. 

As an alternative approach, there could be a new Scene Builder build artifact, a MSI file for per-machine installs. Usually per-machine installs require administrative privileges on Windows whereas per-user installs don't.

The per-machine installer will use Windows default programs folder for installation (`C:\Program Files\SceneBuilder`) whereas the existing per-user installer will use `C:\Users\<UserName>\AppData\Local\SceneBuilder\` instead.

Having 2 artifacts allows administrator to implement those into existing provosioning and deployment processes without additional use of command line flags or switches.

#### Per-User Installer

grafik

#### Per-Machine Installer

grafik

### Issue

Fixes #672 

### Progress


abhinayagarwal:
[Single Package Authoring](http://msdn.microsoft.com/en-us/library/windows/desktop/dd408068(v=vs.85).aspx) has been available for a very long time now. May be we can [override WIX configuration](https://docs.oracle.com/en/java/javase/21/jpackage/override-jpackage-resources.html#GUID-62412365-821D-4736-A20D-70F2D47136D8) to add this.

Oliver-Loeffler:
I am closing this one. Indeed, reworking the packaging to allow a dual purpose MSI file being build is a better approach.

PR #802, opened 2025-02-17
build: use latest release to avoid workflow failure

### LINKED ISSUE #801: build: Java version not found
Early Access workflow failed to run failing in the Setup Java step:

```
Exception in thread "main" java.lang.Error: Could not find URI of JDK[release=23, version=23.0.1, os=linux, arch=x64, type=tar.gz]
	at Download.lambda$main$1(Download.java:77)
	at java.base/java.util.Optional.orElseThrow(Optional.java:403)
	at Download.main(Download.java:77)
	at Download.main(Download.java:42)
```

Looking into the [key-value pairs mapping](https://github.com/oracle-actions/setup-java/blob/main/jdk.java.net-uri.properties) it turns out that `23.0.1` is no longer provided by `jdk.java.net`. We can start using `23.0.2` instead. But this fix can result in future failures when 23.0.2 is also deprecated. I suggest we migrate to using `latest` instead.

### Expected Behavior

Workflows should not fail due to unavailability of Java.

### Current Behavior

Workflows are failing as java version is no longer available for download.

### Steps to Reproduce

Run any worfklow

### Your Environment

N/A

### Screenshots

N/A

abhinayagarwal:
A test run is available [here](https://github.com/abhinayagarwal/scenebuilder/actions/runs/13366022942) which builds with 23.0.2 (latest) released GA version.

### Issue

Fixes #801 

### Progress

### DIFF
--- a/.github/workflows/early-access.yml
+++ b/.github/workflows/early-access.yml
@@ -7,7 +7,7 @@ on:
 
 env:
   JAVA_RELEASE: '23'
-  JAVA_VERSION: '23.0.1'
+  JAVA_VERSION: 'latest'
   JAVAFX_VERSION: '23.0.1'
 
 jobs:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,7 +16,7 @@ on:
 
 env:
   JAVA_RELEASE: '23'
-  JAVA_VERSION: '23.0.1'
+  JAVA_VERSION: 'latest'
   JAVAFX_VERSION: '23.0.1'
 
 jobs:


.github/workflows/early-access.yml
```diff
@@ -7,7 +7,7 @@ on:
 
 env:
   JAVA_RELEASE: '23'
-  JAVA_VERSION: '23.0.1'
+  JAVA_VERSION: 'latest'
```
Oliver-Loeffler, line 10: This is fine. However, technically we could remove the env variable 'JAVA_VERSION' as it is no longer needed by the action. Unfortunately, instead of 'JAVA_RELEASE' the value of 'JAVA_VERSION' is used throughout the whole script. For the `oracle-actions/setup-java@v1.4` the only relevant parameter is `release`.

.github/workflows/release.yml
```diff
@@ -16,7 +16,7 @@ on:
 
 env:
   JAVA_RELEASE: '23'
```
Oliver-Loeffler, line 18: Only one of `JAVA_RELEASE` or `JAVA_VERSION` is effectively needed when we decide to use the latest version. But this could be another PR.

Oliver-Loeffler (approved):
I am fine with this change, as it is nicely small.

PR #807, opened 2025-03-26
Remove default Java version from bundles jobs

### LINKED ISSUE #806: build: Remove Java/JavaFX default versions from bundle jobs
Currently the Java/JavaFX versions are repeated in the different build jobs, as default values, but we always pass the used value from the release job.
To avoid duplication and the need to update on every release, these default values should be removed and the variables should be required, not optional.

jperedadnr:
### Issue

Fixes #806 

### Progress


.github/workflows/release.yml
abhinayagarwal, line 1: Don't we need similar changes in `early-access.yml` ?

abhinayagarwal (approved):

Issue #830, opened 2025-09-06
Scene Builder's Window do not show up


przemSzer:
I try to start Scene Builder, but its window is not showing.

### Expected Behavior

Window should open and I should be able to use scene builder.

### Current Behavior

Window is not visible. 

In Task Manager I can see a process:

<img width="1693" height="332" alt="Image" src="https://github.com/user-attachments/assets/2ed68e9f-b926-4e5d-8b43-99139fb2807e" />

Even if I kill all process and start SB again, the same issue occurres.

### Steps to Reproduce

Double click, or start Scene Builder.

### Your Environment

Windows 11 Proffessional.

The application was working well through few hours (I opened it many times). 
I'm not sure if it is related, but I changed settings in Windows so that *.fxml files should be opened by SceneBuilder.exe. After that, I opened and closed it few times, using double click on fxml file and  the SB worked well. 
Unfortunatelly after some time I wanted to open application again, but window did not show up.

I attach thread dumps taken by Visual VM.

[threaddump-1757147065772.txt](https://github.com/user-attachments/files/22185529/threaddump-1757147065772.txt)
[threaddump-1757147076957.txt](https://github.com/user-attachments/files/22185531/threaddump-1757147076957.txt)
[threaddump-1757145535759.txt](https://github.com/user-attachments/files/22185530/threaddump-1757145535759.txt)

PR #852, opened 2026-03-26
perf: Improve startup workflow and code

Maran23:
This PR improves the startup workflow and code. The goal is to have a more reliable, stable and faster startup.
This aims to fix:
- https://github.com/gluonhq/scenebuilder/issues/846
- https://github.com/gluonhq/scenebuilder/issues/830

What changed
-
`SceneBuilderApp` has now two public static methods:
- `startInBackground(..)`, which takes an action and a task it should do after the action did run. 
This especially helpful to write code like: `startInBackground("Load Data", this::loadData, this::setData);`
Another advantage: The `SceneBuilderApp` is the central point to coordinate such background tasks. So in the future, we can easily rely on that (e.g. to show a `ScrollBar` with all active backgrounds tasks)

Startup code now uses this method to do everything that makes sense in the background, and everything UI related right after. This also avoids any sort of weird race conditions. The process what to do first, next and on the UI Thread is now very well defined.

I also found out that Scenebuilder sometimes processes an empty path for CSS in `PreferencesRecordDocument`. 
In `SCENE_STYLE_SHEETS`, an empty string might be written, which is then read out. This results in `""` to be treated as stylesheet, which will then be converted to your workspace/installation path. The CSS Parser later on then reads all files that are in the root of the Scenebuilder, so for example `.editorconfig`. This leads to a slower startup and CSS errors in the console.

Very minor: The two Scenebuilder images are now created just once in `AppSettings`. In my profiling, loading this images always for a new window made 2%, which is low, but still nice to save.

### Progress

### DIFF
--- a/app/src/main/java/com/oracle/javafx/scenebuilder/app/util/AppSettings.java
+++ b/app/src/main/java/com/oracle/javafx/scenebuilder/app/util/AppSettings.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016, 2024, Gluon and/or its affiliates.
+ * Copyright (c) 2016, 2026, Gluon and/or its affiliates.
  * All rights reserved. Use is subject to license terms.
  *
  * This file is available and licensed under the following license:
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/oracle/javafx/scenebuilder/app/util/AppSettings.java
```diff
@@ -1,209 +1,207 @@
-/*
```
Maran23, line 1: This file was `CRLF` for me, changed it to `LF`. 
You need to click `Hide Whitespace` here on Github to see the changes (only the `Image`).

